PR #13, opened 2022-03-14
bug fix

TheOneAndOnlyDanSan:
Fixed the bug that when shift and space where being press at the same time the player would go up


hashalite:
It would be a bit cleaner if you changed it to:
```
if (freeCamera.input.jumping) {
        velocityY += vSpeed;
}
if (freeCamera.input.sneaking) {
        velocityY -= vSpeed;
}
```
I'll merge if you make this modification.

TheOneAndOnlyDanSan:
i didn't think of that i'll do it

PR #41, opened 2022-07-25
Chinese translation

GodGun968:


hashalite:
Thanks for translating!

hashalite:
@GodGun968 Hello, I have added some new entries in the language file. Would you mind updating zh_cn?

PR #57, opened 2022-12-16
Update zh_cn.json

GodGun968:
It's good that the untranslated text is automatically filled in. This saves me a lot of work.


hashalite:
Thank you!

hashalite:
@GodGun968 could you also check out the language file for the modrinth branch? It has a few strings that are different from the main branch. It is available [here](https://github.com/hashalite/Freecam/blob/modrinth-1.19/src/main/resources/assets/freecam/lang/zh_cn.json).

GodGun968:
Oh well, but why don't you just make one language file?

hashalite:
The modrinth version has some strings that are not used in the main version, so it wouldn’t make sense to include them for both.

GodGun968:
Oh, modrinth's mod is a beta right?

hashalite:
What do you mean?

GodGun968:
I mean the version on modrinth is an updated version right?

GodGun968:
my meaning may not be very clear. forget about it, It doesn't have much to do with me.

hashalite:
The version on Modrinth disables noClip, allowInteract, and freezePlayer while in multiplayer in order to comply with Modrinths rules. The main Freecam mod is considered to be a cheat by Modrinth, and cannot be uploaded there.

GodGun968:
Thanks for the explanation, I've got it.

hashalite:
No problem, and I appreciate your work on the translations!

GodGun968:
😊

PR #67, opened 2023-01-03
Increase Gradle's memory to 2GB

MattSturgeon:
I was unable to build on a fresh machine with only 1GB. Was getting out-of-memory errors while downloading dependencies from maven. Guessing at some point the dependencies went over 1GB without you noticing since you probably have them cached. ¯\\\_(ツ)\_/¯


hashalite:
Thanks!

PR #68, opened 2023-01-04
Use the "swimming" pose in freecam to improve navigation without noClip

MattSturgeon:
The "swimming" pose is normally used when the player is crawling through a one-block-high space. Or when the player activates the swimming hotkey in water. The swimming pose has the smallest player hitbox available in survival, so when `noClip` is disabled the swimming pose is required to get through one-block-high gaps and tunnels.

I figured there's probably zero reason to ever need the larger poses while in freecam, so I've made it so that the `FreeCamera` entity's pose is always set to "Swimming".

We _could_ make the Freecam pose a config option, but I feel like that would needlessly clutter the config menu for little benefit...

Because the eye-hight is matched up with the player's seamlessly, you cannot actually tell your pose has changed until you start moving around and colliding with surfaces.

The new `CameraMixin` fixes a minor issue I found where setting the pose and moving the camera-entity's y-position confused the camera and caused it to show the camera offset by the eye-height difference and slowly transition back to the correct position. The mixin just sets the camera y-position manually whenever the focused camera-entity toggles to or from a `FreeCamera` instance.

I originally noticed this issue when you [pointed out](https://github.com/hashalite/Freecam/pull/25#issuecomment-1127031104) the no-clip complications over on #25, because `FreeCamera` using the "standing" pose causes it to collide really easily with the floor the player is standing on.

### DIFF
--- a/src/main/java/net/xolt/freecam/Freecam.java
+++ b/src/main/java/net/xolt/freecam/Freecam.java
@@ -167,7 +167,7 @@ public class Freecam implements ClientModInitializer {
 
     private static void onEnableFreecam() {
         onEnable();
-        freeCamera = new FreeCamera(-420);
+        freeCamera = new FreeCamera(-420, FreecamPosition.getSwimmingPosition(MC.player));
         freeCamera.spawn();
         MC.setCameraEntity(freeCamera);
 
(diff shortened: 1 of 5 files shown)


hashalite:
Thanks for the PR! Sorry it took me so long to get to these, I recently took a short break from development.

PR #69, opened 2023-01-05
"Initial Perspective" config option

### LINKED ISSUE #16: Create a Mode or Toggle to change the initial position of freecam when activated

MattSturgeon:
Updated replacement for #25. Fixes #16.

- Rebased on main
- Cleaned up and refactored new code
- Now integrates with `FreecamPosition` instead of introducing a new position utils class
- Resolved issues raised [here](https://github.com/hashalite/Freecam/pull/25#issuecomment-1127031104)
  - Fixed issues with No Clip compatibility
  - Moved First Person further forward, to avoid clipping with the player's head
  - Re-arranged settings and changed defaults (as requested)

This works best when combined with #68, since currently the freecam uses "standing" pose. This means that anytime the camera is pointed upward and "Third Person" perspective is selected, the freecam's bounding box will usually collide with the ground.

In cases where we cannot find a non-colliding perspective distance, an error message is shown instead of the usual "Freecam has been enabled" message.

----

### Original description

As requested in #16, I've added an option to select from one of the following initial camera perspectives:

- "Inside" this option retains the current behavior.
- "First Person" this option shifts the camera 0.3 blocks forward, to be just in front of the player's eyes.
- "Third Person" this option replicates F5 mode, moving the camera 4 blocks behind the player.
- "Mirror" this option replicates mirrored F5 mode, moving the camera 4 blocks in front of the player and inverting the camera.

Default is set to "Third Person" as this felt the most intuitive option during testing, but let me know if you'd rather something different and I can amend the commit.

Most of the backed logic comes from `net.minecraft.client.render.Camera`, which has been cleaned up and ported into a new util class `PositionRotationUtil`. Let me know if there's anything you'd like named/formatted differently or if you'd rather a different approach be taken somewhere.

<details><summary><strong>Videos</strong></summary>

https://user-images.githubusercontent.com/5046562/168478005-f961f7cf-6e6d-4992-989c-ba0b08841c78.mov

https://user-images.githubusercontent.com/5046562/168478356-ba94f2ae-80a0-4438-ad72-8863f5fe83ac.mov

https://user-images.githubusercontent.com/5046562/168478145-3baa1d31-a1fe-4fe5-8b98-ffcb821383ab.mov

https://user-images.githubusercontent.com/5046562/168478222-a615f646-4123-45e7-ad21-6a2d8ad15df3.mov

https://user-images.githubusercontent.com/5046562/168478285-b20c59d7-8b34-40a6-a1bc-fbe08a73389b.mov

</details>


MattSturgeon:
I've fixed an number of issues and squashed the PR back down to a single commit. For reference my unsquashed commit history is [here](https://github.com/hashalite/Freecam/commit/504f2c1f49f58e000c9416919c3ae084c179d627).

I've added an additional config option; "Always Check Collision: Whether to check for initial collision, even when using No Clip.". If this option is enabled **or** No Clip is disabled, then collision is checked.

When checking collision, [`FreeCamera.moveForwardUntilCollision()`](https://github.com/hashalite/Freecam/pull/69/files#diff-de0f8c8ad4ae6e5f24247dc63ad9c0ddc3ab4891dfa05f16ebb9bbad0ba114b5R83) gradually moves the `FreeCamera` position toward the configured Initial Perspective, stopping either when that perspective is reached **or** when the `FreeCamera` would collide if moved any further. Currently it moves [`0.1` blocks at a time](https://github.com/hashalite/Freecam/pull/69/files#diff-de0f8c8ad4ae6e5f24247dc63ad9c0ddc3ab4891dfa05f16ebb9bbad0ba114b5R96), but this increment could be reduced at the cost of performance, if preferred.

If collision is being checked and the `FreeCamera` is unable to move at all without colliding, an error is shown to the user in the status text: "Unable to use camera perspective without clipping into blocks.` Unless #68 is merged, this will happen a lot, because the `FreeCamera`'s feet collide with blocks in the "Standing" pose.

MattSturgeon:
Tested No Clip disabled with #68 [merged](https://github.com/MattSturgeon/Freecam/tree/initial-position-swim-pose):

https://user-images.githubusercontent.com/5046562/211163849-28781d20-b777-4061-8a5e-04428136e315.mp4

As you can see, Third Person perspective works as intended and the camera is able to access any area the player could using elytra/swimming/trapdoor-trick/etc methods.

And without it merged:

https://user-images.githubusercontent.com/5046562/211163845-48401c23-bbf0-4a7a-ad71-6a1250fb3fc4.mp4

As you can see, it works intuitively when merged. When not merged, the Third Person perspective only works properly when pitch > 0. Conversely, First Person perspective only works when pitch < 0.

Issue #55, opened 2022-11-01
[Enhancment] clipping freecam can fly through transparent blocks


BeeFox-sys:
free cam being able to fly though blocks like glass would be very nice. it would still prevent xray for players who do not want that, but allow to leave structures provided that there are windows to leave through

PR #71, opened 2023-01-07
Add new "Ignore Collission" settings

MattSturgeon:
Adds a new "Collision Options" sub-menu to the mod config menu. The existing "No Clip" and "Always Check Collision" options are renamed and moved into this sub-menu.

Additionally, two new options are added: "Ignore Transparent Blocks" and "Ignore Openable Blocks".

- Adds a new "Ignore Transparent Blocks" option
- Adds a new "Ignore Openable Blocks" option, to allow traveling through doors (etc).
- Replaces "No Clip" with the "Ignore All Collision" option.
- Moves "Always Check Collision" into the sub-menu.
- Introduced a new `CollisionWhitelist` class which decides whether blocks are transparent or openable.

https://user-images.githubusercontent.com/5046562/223201911-bd539ac8-c668-4bdd-950d-fa33e3c1354f.mp4

<details>
<summary><strong>Original description</strong></summary>

Adds a new "Opaque" collision mode, requested in #55. This mode allows `FreeCamera` to clip through any material that doesn't block light (`Material.blocksLight() == false`).

Replaces No Clip with the "None" mode. The "All" mode functions the same as disabling No Clip.

If `Material.blocksLight()` leads to false-positives we could manually curate a transparent block list or find some alternative method of deciding. In my testing it worked intuitively. _EDIT: some light emitting blocks seem to be false positives... going with a whitelist._

I've removed the No Clip translation strings, but I've only added English strings for Collision Mode as I don't speak Italian or Chinese.

This PR will likely have merge conflicts with my other recent PRs, so I'll have to rebase whichever stays open longer. _EDIT: done._

</details>

Fixes https://github.com/hashalite/Freecam/issues/55


MattSturgeon:
Not sure if it's best to have "Collision Mode" configuring which blocks _are_ collided with, or "Clipping Mode" configuring which blocks _aren't_... Or maybe there's another alternative name?

If inverted to "Clipping Mode" then "Opaque" would be renamed to "Transparent", which _might_ be more intuitive... Maybe?

MattSturgeon:
I'll try and rebase this later. I think I'll have to rethink some things so this doesn't interfere with _initial perspective_'s collision checks.

MattSturgeon:
Rebased, still need to investigate Initial Perspective compatibility...

MattSturgeon:
@hashalite ready for review.

- Fixed compatibility with "Initial Perspective", made it so that "Always Check Collision" only overrides "ignore all", rather than also overriding "ignore transparent"... thoughts?
- Replaced the `Material.blocksLight()` check with `CollisionWhitelist.isTransparent(Block)` since I was getting false-positives with light-emitting blocks, etc.
- Added an additional mode "Doors, etc" (`IGNORE_OPENABLE`) which ignores collision of all doors, trapdoors and fencegates.

Tempted to add an `IGNORE_TRANSPARENT_OR_OPENABLE` but worried that'd start to clutter the config menu... thought?

MattSturgeon:
Decided it makes more sense to separate each "mode" into its own boolean option, that way they can be used simultaneously. Additionally it means each option can have its own tooltip. There's also less clutter at the top of the config menu.

Considering whether `ignoreAllCollision` and `checkCollision` should be combined into an enum? e.g. {`INIT`,  `NEVER`, `OFF`}?

<details>
<summary>Screenshots</summary>

transparent tooltip
openable tooltip
all tooltip

The original version, for comparison:

Enum config option



</details>

MattSturgeon:
Finally come up with a config that makes some sense to me. Used `@CollapsibleObject` to make a "Collision Settings" sub-menu: 

https://user-images.githubusercontent.com/5046562/222915642-dc428991-8cc9-4e3d-b7bf-74b9c53c56fe.mp4

Also squashed everything down, went over the option names and tooltips and updated the commit message.

Ideally, we'd have some collision options disabled automatically when they are redundant, but without https://github.com/shedaniel/cloth-config/issues/26 there isn't a simple way to achieve that.

hashalite:
This looks great, however I am kind of reluctant to merge right now because of the changes I would have to make to the Modrinth branch. In order to comply with Modrinth's rules, it disables NoClip whilst in multiplayer, and I'm not yet sure how I would handle these new settings when merging. I could just disable all of them when in multiplayer, but there is a chance that clipping through transparent blocks fits within the rules. I would have to check with Modrinth's team to be sure.

MattSturgeon:
> In order to comply with Modrinth's rules, it disables NoClip whilst in multiplayer, and I'm not yet sure how I would handle these new settings when merging.

Disabling all of the ignore collision option would be the safe option.

> There is a chance that clipping through transparent blocks fits within the rules. I would have to check with Modrinth's team to be sure.

👍 Sounds good. Can always enable specific options later on if they're ok with them.

src/main/java/net/xolt/freecam/mixins/BlockCollisionSpliteratorMixin.java
```diff
@@ -0,0 +1,57 @@
...
+    // Store the provided entity, so we can later check if it is freecam
+    @Inject(method = "<init>(Lnet/minecraft/world/CollisionView;Lnet/minecraft/entity/Entity;Lnet/minecraft/util/math/Box;Z)V", at = @At("RETURN"))
+    private void onConstructed(CollisionView world, Entity entity, Box box, boolean forEntity, CallbackInfo ci) {
+        this.entity = entity;
+    }
+
+    // Apply custom block collision rules to freecam
+    @Redirect(method = "computeNext()Lnet/minecraft/util/shape/VoxelShape;", at = @At(value = "INVOKE", target = "Lnet/minecraft/block/BlockState;getCollisionShape(Lnet/minecraft/world/BlockView;Lnet/minecraft/util/math/BlockPos;Lnet/minecraft/block/ShapeContext;)Lnet/minecraft/util/shape/VoxelShape;"))
+    private VoxelShape onGetCollisionShape(BlockState blockState, BlockView world, BlockPos blockPos, ShapeContext context) {
+        if (entity instanceof FreeCamera) {
```
MattSturgeon, line 37: Can add an early check here to apply, or could add the checks to specific cases if allowed.

```suggestion
        if (entity instanceof FreeCamera && Freecam.canUseCheats()) {
```

src/main/java/net/xolt/freecam/mixins/BlockCollisionSpliteratorMixin.java
```diff
@@ -0,0 +1,57 @@
...
+            // Unless "Always Check Initial Collision" is on and Freecam isn't enabled yet
+            if (!ModConfig.INSTANCE.collision.alwaysCheck || Freecam.isEnabled()) {
+                // Ignore all collisions
+                if (ModConfig.INSTANCE.collision.ignoreAll) {
```
MattSturgeon, line 41: If modrinth say we can use `ignoreTransparent` and `ignoreOpenable` in multiplayer, we can move the check to `ignoreAll` specifically.

```suggestion
                if (ModConfig.INSTANCE.collision.ignoreAll && Freecam.canUseCheats()) {
```

MattSturgeon (commented):
Proposed changes for Modrinth compliance

MattSturgeon:
Is integration with the modrinth branch the only thing blocking this PR? 

I don't see the issue with maintaining the current behaviour in the modrinth/multiplayer scenario. As per [this suggestion](https://github.com/hashalite/Freecam/pull/71#discussion_r1125563800).

We won't have _lost_ anything by also disabling the new modes if we're unsure of their compliance.

If we later discover that some modes do comply, that would be a simple PR to enable them.

Let me know if you have any other concerns 🙂

hashalite:
I was a bit concerned with how cluttered the config screen is getting, but on second thought I think the collapsible category looks good enough. I will go ahead and merge. Thanks again!

MattSturgeon:
As mentioned [in this comment](https://github.com/hashalite/Freecam/pull/71#issuecomment-1482140393), the config screen is starting to get a little busy and if we're not careful it could soon be overcrowded and confusing.

This inspired me to begin work on https://github.com/shedaniel/cloth-config/pull/195, but in the meantime it might be a good idea to consider how the current config could be improved.

Perhaps we could make use of tabbed category pages, collapsible sub-categoried (either expanded or collapsed by default) and other config entry types?

In the future we could also make use of dependencies to hide/disable some options when relevant.

PR #85, opened 2023-04-01
Option to remember F5 state

MattSturgeon:
I think most users would expect F5 state to persist when toggling freecam. This could either be fixed behaviour (no config option), a [hidden config option](https://shedaniel.gitbook.io/cloth-config/auto-config/annotations#configentry.gui.excluded) (only accessible via config files) or a normal config option.

In this PR I made the assumption that a bunch of existing config options might make sense bundled together into a "toggle" sub-category, but I'm open to alternative suggestions.

- Introduced "Remember F5" which re-sets the previous F5 perspective after exiting freecam
- Moved initial perspective into toggle sub-category.
- Moved disableOnDamage into toggle sub-category
- Moved freecam & tripod notifications into toggle sub-category

Not sure if all of the options I moved to the toggle sub-category make sense there. For example it might make sense to have a separate "notifications" category, especially if more things will be notified in the future.

Needed to add an `onDisabled()` method (called immediately after disabling), otherwise `GameOptionsMixin` prevents setting F5 state. I didn't add an `onEnabled()` because it'd currently be empty. I didn't look into whether the current `onEnable()` and `onDisable()` could be changed into `onEnabled()` and `onDisabled()` instead of adding additional methods - not sure if anything would break when calling those _after_ toggling...


hashalite:
Thank you!

PR #115, opened 2023-11-10
I hacked in a spectator-like follow mode

jpicht:
Is this kind of thing something that you'd consider to merge? I'd be willing to put in some work to make it nicer, this is just a simple hack that _works for me_:tm:

### DIFF
--- a/src/main/java/net/xolt/freecam/Freecam.java
+++ b/src/main/java/net/xolt/freecam/Freecam.java
@@ -4,14 +4,19 @@ import me.shedaniel.autoconfig.AutoConfig;
 import net.fabricmc.api.ClientModInitializer;
 import net.fabricmc.fabric.api.client.event.lifecycle.v1.ClientTickEvents;
 import net.fabricmc.fabric.api.client.keybinding.v1.KeyBindingHelper;
+import net.fabricmc.fabric.api.object.builder.v1.entity.FabricEntityTypeBuilder.Living;
 import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.input.Input;
 import net.minecraft.client.input.KeyboardInput;
 import net.minecraft.client.option.KeyBinding;
 import net.minecraft.client.option.Perspective;
 import net.minecraft.client.util.InputUtil;
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.LivingEntity;
 import net.minecraft.text.Text;
+import net.minecraft.util.hit.EntityHitResult;
 import net.minecraft.util.math.ChunkPos;
+import net.minecraft.util.math.Vec3d;
 import net.xolt.freecam.config.ModConfig;
 import net.xolt.freecam.util.FreeCamera;
 import net.xolt.freecam.util.FreecamPosition;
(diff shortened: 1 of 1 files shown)


src/main/java/net/xolt/freecam/Freecam.java
```diff
@@ -27,12 +32,16 @@ public class Freecam implements ClientModInitializer {
     private static KeyBinding playerControlBind;
     private static KeyBinding tripodResetBind;
     private static KeyBinding configGuiBind;
+    private static KeyBinding followBind;
+    private static KeyBinding setFollowBind;
     private static boolean freecamEnabled = false;
     private static boolean tripodEnabled = false;
+    private static boolean followEnabled = false;
```
MattSturgeon, line 39: The `isEnabled` method should also check this. Currently various mixins are broken, e.g. "Show Player".

src/main/java/net/xolt/freecam/Freecam.java
```diff
@@ -82,6 +105,54 @@ public void onInitializeClient() {
...
+        }
+
+        if (!(lookedAtEntity instanceof LivingEntity follow)) {
+            return null;
+        }
+
+        return follow;
+    }
+
+    public static void toggleFollow() {
```
MattSturgeon, line 125: `toggleFollow` might benefit from being split into `enableFollow` and `disableFollow` methods?

src/main/java/net/xolt/freecam/Freecam.java
```diff
@@ -181,8 +252,17 @@ private static void onDisableTripod() {
...
+        if (!followEnabled) {
+            onEnable();
+        }
+
         freeCamera = new FreeCamera(-420);
+
+        if (followEnabled) {
+            freeCamera.applyPosition(new FreecamPosition(followMe));
+            followEnabled = false;
+        }
```
MattSturgeon, line 264: This should use `FreecamPosition.getSwimmingPosition()` instead of `new FreecamPosition()`.

It may be cleaner to remove the `public FreeCamera(int id)` constructor entirely and instead do something like:

```java
freeCamera = new FreeCamera(-420, FreecamPosition.getSwimmingPosition(
    followEnabled ? followMe : MC.player
));

MattSturgeon: Also, calling `onEnable()` conditionally and then manually disabling `followEnabled` without any sorta `disableFollow` method feels a bit messy.

MattSturgeon (changes requested):
Funnily enough, I was just about to look into working on this feature myself, before I found your PR. :joy: 

I've given it a quick test and had a quick glance over the code. It's working very well as an initial proof of concept.

Other than code cleanup/refactoring, I personally would take a different approach to the keybinding. IMO it makes sense to re-use vanilla's "Pick" keybind (middle mouse) rather than adding a new freecam-specific one.

This would mean "follow" is always activated from within freecam mode, and we'd have to think about how to exit follow mode; perhaps the freecam keybind or using the "pick" bind would exit?

Even if we don't go with my approach, I still think the "follow" and "followSet" binds should be combined somehow.

Hopefully @hashalite is happy to include a follow/spectate mode. If I get chance I may look into working on this myself too, although I have other stuff I want to look at too, so who knows.

PR #121, opened 2023-12-03
Minor config GUI improvements

MattSturgeon:
Two relatively minor improvements to the config GUI.

Other than speed scale, none of these affect how config is stored, saved or loaded. Out of bounds numbers are still permidded and disabled options can still be modified in the config file or by modifying and saving in code. Only the GUI widget is actually disabled as a visual cue.

## Collision mode requirements

Taking advantage of the experimental Cloth Config requirements API added in shedaniel/cloth-config#204,  "Ignore Transparent" and "Ignore Openable" are now _disabled_ when "Ignore All" is switched "ON".

### Alternative approach

Rather than _disabling_ "Ignore Transparent" and "Ignore Openable", we could also _hide_ them instead. This would probably need some text added to hint at the hidden options.

We could also move them into a nested sub-category and disable _that_. The current implementation is the simplest option, but I'm open to doing things differently.

### Implementation note

It's technically a bit hacky to the AutoConfig transformer system for this;

The main issue you might run into is run order: another transformer could run after ours, which could decide to remove/replaced the widget we've captured, rather than mutating it.

Luckily AutoConfig will never do that to boolean entries and we won't either. So for us it's a non-issue.   
  _Incidentally, that edge-case is why I opened shedaniel/cloth-config#215._

I've left a FIXME note as a reminder anyway. If I ever finish a Requirements API for AutoConfig, we can move over to that once its merged.

## Movement speed sliders

Moved to #190 

<details><summary>Details</summary>
<p>

Replace the text-entry widgets with number slider widgets.

This meant I had to a) set an upper bound on the value and b) convert to a scaled up `int`.

Some users will still have the old scaled-down values and they will move very slowly unless they reset/adjust their settings after upgrading.

We could avoid that by renaming the fields, however the current names are pretty good (`horizontalSpeed` & `verticalSpeed`)...

</p>
</details> 

## Minor Enum cleanup

I almost pushed this to #122, but figured this PR was more topical. This minor tweak was made while working on #116, but felt unrelated to the core change of that PR. 

## Demo

https://github.com/hashalite/Freecam/assets/5046562/d88de289-9971-4d82-ab1b-74b2d406c11f

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and Freecam's versioning is based on [Semantic Versioning](https://semver.org/sp
 ### Changed
 
 - Movement speed options now use sliders instead of text fields ([#190](https://github.com/MinecraftFreecam/Freecam/pull/190)).
+- Redundant collision options are now dynamically hidden ([#121](https://github.com/MinecraftFreecam/Freecam/pull/121)).
 
 ### Removed
 
--- a/common/src/main/java/net/xolt/freecam/config/ModConfig.java
+++ b/common/src/main/java/net/xolt/freecam/config/ModConfig.java
@@ -121,49 +121,52 @@ public class ModConfig implements ConfigData {
     }
 
     public enum FlightMode implements SelectionListEntry.Translatable {
-        CREATIVE("text.autoconfig.freecam.option.movement.flightMode.creative"),
-        DEFAULT("text.autoconfig.freecam.option.movement.flightMode.default");
+        CREATIVE("creative"),
+        DEFAULT("default");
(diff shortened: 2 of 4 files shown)


MattSturgeon:
This PR will likely be redundant because we need to move away from autoconfig to (cleanly) include the modrinth behaviour in the main branch.

MattSturgeon:
I've rebased this on main.

I think we need to think about how/if the setting requirements also make sense on the Modrinth Edition.

Also, a changelog entry should be added.

hashalite (approved):

PR #122, opened 2023-12-03
Minor build system improvements

MattSturgeon:
<details><summary>Original description</summary>
<p>

This PR contains a collection of small commits that each didn't warrant their own PR.

Feel free to cherry-pick what you like, or request changes/removals here.

Probably best to review each commit individually, as the changes are mostly unrelated to each other.  
Each commit also has brief notes in the commit message.

</p>
</details> 

Minor improvements to the build system:

- ~~Workaround https://github.com/architectury/architectury-templates/issues/28~~
- Bump gradle to 8.5
- Use gradle's "all" distribution to enable gradle API javadocs
- Disable generating IDE run configs for the server
- Improve IDE run config names
- Fix jar archive names
  - base name is project name ("freecam") plus path (":fabric"), replacing ":" with "-".
  - archive version is project version, "+" & minecraft version.
  - this matches previous releases. 
- Add [Devlogin](https://github.com/PlanetTeamSpeakk/DevLogin) support to `localRuntime`.
  - allows login into a MS Account to debug things on non-"cracked" multiplayer servers.
  - can be used via "Run logged in" run configs or `runLoggedIn` gradle tasks.
- Set `accessWidenerPath` once, in a central location.
- Set `shadow` plugin version once, in a central location.

PR #123, opened 2023-12-07
Refactor build system to support modrinth tweaks

MattSturgeon:
## Status

This PR is finished, however it isn't the best way to approach the issue. See [this comment](https://github.com/hashalite/Freecam/pull/123#issuecomment-1856122710).

The replacement will involve moving **away from Autoconfig** & using [Java SPI](https://docs.oracle.com/javase/tutorial/sound/SPI-intro.html) to load **different [`Behaviour`](https://github.com/MattSturgeon/Freecam/blob/modrinth/common/src/main/java/net/xolt/freecam/config/Behaviour.java) implementations.**

I'll likely close this in favor of another PR once I have one ready.

---

## Original description

I'm opening this PR early in an unfinished state to give an idea of what I'm working on for integrating the modrinth tweaks into the main branch.

At this stage I've stubbed out a gradle multi-project build as follows:

```
> Task :projects

------------------------------------------------------------
Root project 'freecam'
------------------------------------------------------------

Root project 'freecam'
+--- Project ':common'
+--- Project ':fabric'
+--- Project ':fabric-modrinth'
+--- Project ':forge'
+--- Project ':forge-modrinth'
\--- Project ':modrinth'
```

TODO:

- [x] Refactor `allprojects` and `subprojects` usage into ["convention" plugins](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins)
- [x] Refactor repeated config into convention plugins
- [x] Add additional `:modrinth` project which uses the same sources as `:common`
- [x] Add `-modrinth` varients of `:forge` and `:fabric` projects, using `:modrinth` instead of `:common`
- [x] Inject additional properties in `processResources` to allow "name" to be "Freecam (Modrinth Edition)"
- [x] Ensure jars are named correctly (e.g. `freecam-fabric.jar`) _(see #122)_
- [x] Implement a JSON overloading task to add/replace `lang` file entries
- [x] Generate a "`BuildConfig`" class which contains static-final constants such as `CHEATS_RESTRICTED` and `TOOLTIP_COUNTS`
- [x] Merge the actual changes from the [modrinth branch](https://github.com/hashalite/Freecam/compare/main...modrinth-1.20)
- [ ] Final cleanup of git history (after code review & other PRs being merged)

<details><summary>Preprocessing</summary>
<p>

~~Use [JCP](https://github.com/raydac/java-comment-preprocessor) or [RM's equivalent](https://github.com/ReplayMod/preprocessor) to pre-process the common/modrinth sources~~ _EDIT: I've done a POC using JCP, which reminded me how clunky java preprocessing can be. I think generating a config class would be simpler:_

- Dependency injection - `:common` & `:modrinth` inject different implementations of some interface
- Source code generation - a class is rendered using a gradle task
- Self-targeting mixins - `:modrinth` adds mixins that hook into `:common`'s code

~~Personally I think preprocessing is the simplest, especially for the case where `@ConfigEntry.Tooltip` annotations need different values.~~ It would mean additional noise where changes are required, however I don't think there will need to be a lot of differences. _EDIT: annotation values can be set to static-final fields, so code-gen could still work here..._

Lang files would need to be handled separately. I'm imagining a gradle copy task that can be used within `processResources`. The task would parse the base lang file as json, then parse an override file as well. Merge the changes and write the result to the build outputs. _EDIT: this is done._

</p>
</details> 

### Note on Convention Plugins

Convention plugins are small gradle plugins located in `buildSrc` that each describe a "trait" of our config. They can be used modularly throughout the actual config as a more flexible alternative to `allprojects` and `subprojects`. The [official guide](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins), and [sample](https://docs.gradle.org/current/samples/sample_convention_plugins.html) [projects](https://docs.gradle.org/current/samples/sample_sharing_convention_plugins_with_build_logic.html) are worth a quick look if you're unfamiliar. After these changes most of the shared configuration logic lives under `buildSrc`.


MattSturgeon:
This is coming along now. Just need to merge the modrinth branch's mixin patches, expand more properties in `mods.toml` & `fabric.mod.json`, cleanup git history, rebase once #122 is merged, test, polish & review.

There's two main systems in play to handle the modrinth differences:

### JSON overloading

I've introduced a small gradle plugin which hooks into `processResources` in order to modify assets

MattSturgeon:
@hashalite This is feature complete, somewhat tested and ready for review. :tada: 

There are some related commits that could be squashed further before merging, and others that would become redundant if other PRs get merged. I've left the history somewhat granular to hopefully aid review.

If anything looks complex I'm happy to explain and/or refactor.

MattSturgeon:
After a conversation with @wagyourtail there's a few improvements that can be made to this PR:

- By switching from autoconfig to cloth-config proper we avoid needing to inject translation keys into the lang files; each key could have a `.modrinth` variant all in one file.
  - This lessens the need for code-gen & JSON merging.
  - This will also give better control of the config GUI & stored config values.
  - This could be done in a separate PR.
- [`ServiceLoader`](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) can be used to inject different implementations of `Behaviour`
  - This removes the need for any code generation.
- (A bit more radical) porting from architectury to [unimined](https://github.com/unimined/unimined) may enable a simpler project layout.
  - I wasn't aware this existed when writing #118 
  - unimined is less prescriptive, and allows more control over using `sourceSets` vs `projects`
  - unimined can still use archi-api, however I've been thinking of dropping this dependency anyway...
  - Potential downsides
    - Less documentation (not that architectury has much)
    - No `migrateMappings` task (not that manual mapping updates are difficult)
    - No `ExpectPlatform` API, would just use `ServiceLoader`s.

MattSturgeon:
Closed in favor of #132

PR #124, opened 2023-12-11
Minor mixin tweaks and cleanup

MattSturgeon:
Moved mixin related changes out of #122 

- Take advantage of Access Widener to avoid needeng to use mixin's `Accessor`s.
- Formatting cleanups
- Fixed ambiguios mixin references
- Marked added members as `Unique` (ensures we don't override vanilla class members, or members added by other mods)
- Improve the collision mixin's code


hashalite:
Thank you again!

PR #126, opened 2023-12-14
General cleanup; keypress handling; tripod storage; FreecamPosition

MattSturgeon:
This PR refactors the tripod storage system, tick handlers, position setting & minor related changes.

User-facing changes:
- Toggle keybind is consumed on release.
  -  More intuitive when held to toggle tripods.
  - Toggles only if it wasn't used to toggle a tripod.
  - Toggles only if it warn't held for over 10 ticks.
  - Still works if pressed for less than one tick.
- Resetting a tripod no longer moves the camera to the player's feet.
- Tripods should get stored separately for modded dimensions.

Internal changes:
- ~~Reduced mixin usage for tick events~~
  - _(merged in #129)_
- Refactored keypress handling in `postTick`
  - Added `reset` to `ModBinding`
  - Updated `ModBinding`'s wrapper method names to new mappings.
  - _(See user-facing changes)_
- Extracted tripod storage to `TripodRegistry`
- Access tripods using `TripodSlot` enum
  - enables using `EnumMap`, which is just an array that implements `Map` and therefore very efficient.
  - removes the need to convert `keyCode` to a number all over the place
- Remove `FreecamPosition.getSwimmingPose()` since we _always_ want the swimming pose
  - This fixed at least one bug (mentioned above) and will likely prevent others
- (in the final commit) I refactored position setting a bit
  - Intended to centralize logic, in particular "initial perspective"
  - Should also help with stuff like #116 
  - This commit could be dropped if not considered necessary


hashalite:
I agree with most of the changes here, however I'm not sure I understand the motivation behind limiting the amount of time the toggle bind can be held. It doesn't feel any more natural to me. Could you explain further?

MattSturgeon:
> I agree with most of the changes here, 

Great!

> I'm not sure I understand the motivation behind limiting the amount of time the toggle bind can be held.

One scenario: you've held it down for a while, thinking about which tripod to open, but you then change your mind.

For me, pressing the key for more than ~0.2s stops feeling like I'm "hitting" the key and starts to feel like "holding". Another analogy is tapping vs holding on touchscreen devices.

0.5s (10 ticks) felt like a reasonable cut-off for distinguished a _slightly_ long tap from a mistakenly held key, but this could be set to anything really.

If you disagree, `TOGGLE_KEY_MAX_TICKS` could be changed or removed entirely.

common/src/main/java/net/xolt/freecam/tripod/TripodSlot.java
```diff
@@ -0,0 +1,35 @@
...
+    public String toString() {
+        return this == NONE ? "None" : String.valueOf(ordinal());
+    }
+
+    public static boolean inRange(int number) {
+        return number >= MIN && number <= MAX;
+    }
+
+    public static TripodSlot valueOf(int number) throws IndexOutOfBoundsException {
+        if (inRange(number)) {
```
MattSturgeon, line 21:
```suggestion
        if (!inRange(number)) {
```

Oops

common/src/main/resources/assets/freecam/lang/en_us.json
```diff
@@ -6,9 +6,9 @@
   "key.freecam.configGui": "Config GUI",
   "msg.freecam.enable": "Freecam has been enabled.",
   "msg.freecam.disable": "Freecam has been disabled.",
-  "msg.freecam.openTripod": "Opening camera #",
-  "msg.freecam.closeTripod": "Closing camera #",
-  "msg.freecam.tripodReset": "Reset camera #",
+  "msg.freecam.openTripod": "Opening camera #%s",
```
MattSturgeon, line 9: Should the `#` be part of `TripodSlot::toString`? In the future you could have non-number tripods which don't need a `#` 🤔

common/src/main/java/net/xolt/freecam/tripod/TripodSlot.java
```diff
@@ -0,0 +1,35 @@
...
+
+public enum TripodSlot {
+    NONE, ONE, TWO, THREE, FOUR, FIVE, SIX, SEVEN, EIGHT, NINE;
+
+    public static final int MIN = 1;
+    public static final int MAX = values().length - 1;
+
+    @Override
+    public String toString() {
+        return this == NONE ? "None" : String.valueOf(ordinal());
```
MattSturgeon, line 13:
```suggestion
        return this == NONE ? "None" : "#%d".formatted(ordinal());
```

Could move the `#` here?

MattSturgeon:
This has been rebased too.

hashalite:
> One scenario: you've held it down for a while, thinking about which tripod to open, but you then change your mind. I think toggling on release here is unintuitive.

This is a good point!

> Another analogy is tapping vs holding on touchscreen devices...

Also a good point.

Everything looks good, will go ahead and merge. Thanks again, all of this code cleanup has been 🤌

PR #129, opened 2023-12-15
Drop requirement on Architectury API

MattSturgeon:
## Motivation

Fewer dependencies tied to minecraft versions/platforms makes it easier to support new/different mc versions without relying on others to do so first.

Although this means we have to manually use the various platform APIs we target, this isn't actually a big deal in our case, since we use very few APIs:
- Tick events
- Init event
- Keybind registration

Using the APIs directly may also give us more control.

## Implementation

- Cherry-picked the event handler refactoring from #126 
- Removed the Architectury API dependency from gradle config
- Implemented Fabric API calls
- Implemented Forge event handlers
- Refactored the mod's keybinding declaration (see below)

## Key bindings

Previously, keybinds were stored as discrete fields on the main `Freecam` class. This meant having to construct a separate `List` (or stream) in order to iterate over them when registering.

Rather than maintain a list of keybinds twice, which would be tedious and prone to error, I've introduced the `ModBindings` enum in the `config` package.

This has several advantages:
- We automatically get a list of all our key binds via `Enum::values`
- We can make declarations more readable using a custom constructor
- We can (more easily) construct the keybinds lazily, ensuring they aren't added to `KeyBinding`'s maps until we're ready to register them.

To minimize boilerplate elsewhere I've wrapped the `isPressed()` and `wasPressed()` methods. We could also introduce our own `wasPressedReset()` to improve handler code, e.g:

```java
public boolean wasPressedReset() {
    if (wasPressed()) {
        // Alternatively, use AW to access KeyBinding::reset
        // get().reset();
        while (wasPressed()) {}
        return true;
    }
    return false;
}
```

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -8,9 +8,6 @@ loom {
 
 dependencies {
     modImplementation "net.fabricmc:fabric-loader:${rootProject.fabric_loader_version}"
-
-    modApi "dev.architectury:architectury:${rootProject.architectury_version}"
-
     modApi "me.shedaniel.cloth:cloth-config:${rootProject.cloth_version}"
 }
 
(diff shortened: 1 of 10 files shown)


hashalite:
Thank you sir!

Issue #130, opened 2023-12-19
Move away from AutoConfig


MattSturgeon:
### Background & motivation

I'm creating this issue to track & discuss a move away from AutoConfig.

AutoConfig is convenient for simple configuration because you can declare your config, add a few annotations, and watch in awe as both a frontend (GUI) and backend (save/load) is created automatically on-the-fly.

The generated GUI uses Cloth Config and the backend (in our case) is based on `JanksonConfigSerializer`.

It does have some limitations though:

- Tooltips must be declared statically using annotations
  - This makes it harder to include the modrinth changes in the main branch
  - I worked around this in #123 by merging lang json files and generating `static final` tooltip counts at build time.
- Widgets are limited and (mostly) chosen automatically
  - You have some control and can register your own config entry providers/transformers, but this starts to become nearly as complex as using Cloth-Config directly.
- [Dependencies](https://github.com/shedaniel/cloth-config/pull/204) aren't easy to add to config entries
  - You [_can_ do it](https://github.com/hashalite/Freecam/pull/121/commits/b8de5bafdbc00e442865f2abc1b2dba3f5971424#diff-6e7c95e626c411cdfd710a1d72350d4f9561307f954c79636e4f58791a9ce01cR48-R56) via registering transformers though.

For an example of using Cloth Config directly, one can read [their demo](https://github.com/shedaniel/cloth-config/blob/v8/common/src/main/java/me/shedaniel/clothconfig2/ClothConfigDemo.java) and browse [their docs](https://shedaniel.gitbook.io/cloth-config/using-cloth-config/creating-a-config-screen).

It's also worth noting that while still maintained and actively updated to new Minecraft versions, [@shedanial](https://github.com/shedaniel) is no longer actively _developing_ Cloth Config (adding new features, reviewing PRs, etc). To quote the FAQ:

> **Are you still maintaining Cloth / Auto Config?**
> Yes, but only to fix bugs and update it to new Minecraft versions
> [...] Cloth / Auto Config is unlikely to receive new features.

He even notes **potential alternatives**, conceding that they may not be quite as good:

> Here are some notable alternatives that I think (personally), are pretty good as well.
> - [MidnightConfig](https://github.com/TeamMidnightDust/MidnightLib/wiki)
> - [owo-lib](https://docs.wispforest.io/owo/config/)
> - [OroConfig](https://github.com/OroArmor/Oro-Config)
> - [Omega Config](https://github.com/Draylar/omega-config)
>
[...]

PR #132, opened 2023-12-19
Refactor build system to support Modrinth tweaks [v2]

MattSturgeon:
After experimenting with this in #123, helpful insight from @wagyourtail which highlighted [fundamental flaws](https://github.com/hashalite/Freecam/pull/123#issuecomment-1856122710) in that initial approach, and plenty of assistance from @Juuxel in understanding loom's quirks, I'm ready to open this draft and begin iterating towards something we can call "done".

Note: until #130 is resolved, I'll avoid implementing the modrinth-specific tooltips.   
Alternatively, we could write a [config entry transformer](https://shedaniel.gitbook.io/cloth-config/auto-config/custom-gui-handlers#guitransformer) which provides a different tooltip depending on the variant.

## TODO

- [x] Build an additional `modrinth` "variant" for each platform.
- [x] Provide a run config for the variant.
- [x] Build different metadata (`mods.toml`, `fabric.mod.json`) for each variant.
- [x] Write a "normal" and "modrinth" service provider.
- [x] Have each variant depend on the appropriate Service Provider implementation.
- [x] Have config tooltips/descriptions be different for each variant:
- [x] Add a `VariantTooltip` annotation where count can be declared for each variant.
- [x] Write a config entry gui transformer similar to the [default `Tooltip` transformer](https://github.com/shedaniel/cloth-config/blob/e216516e8a95249059e71275e7993dae38b61b80/common/src/main/java/me/shedaniel/autoconfig/gui/DefaultGuiTransformers.java#L51-L77).
- [x] Ensure production builds work correctly.
- [ ] Refactor the forge/fabric variant-specific gradle code into a shared "variant" plugin.

This PR supersedes #123.


MattSturgeon:
@hashalite this is ready for review.

The first 6 commits are cherry-picked from #122. This PRs actual changes are from c0e936965fe2167f9c8717c0d60a3ddb6454e508 onward.

I ended up going with a _fairly_ simple `VariantTooltip` AutoConfig GUI transformer so that this PR isn't blocked by #130. This also lessens the need for #130.

I'm a little unhappy with how much duplicated code there is across `forge/build.gradle` & `fabric/build.gradle`. This could be improved by introducing a custom gradle plugin that handles the variant stuff.

That said, the PR is technically complete, so I figured I'd ask for your thoughts & review.

common/src/main/java/net/xolt/freecam/config/VariantTooltip.java
```diff
@@ -0,0 +1,29 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface VariantTooltip {
+
+    Count[] value() default {};
+
+    @Retention(RetentionPolicy.RUNTIME)
+    @interface Count {
+        String variant() default "all";
+        int value() default 0;
```
MattSturgeon, line 27:
```suggestion
        int value() default 1;
```

common/src/main/java/net/xolt/freecam/config/ModConfig.java
```diff
@@ -17,12 +18,14 @@ public class ModConfig implements ConfigData {
...
     public static void init() {
         AutoConfig.register(ModConfig.class, JanksonConfigSerializer::new);
+        ConfigExtensions.init(AutoConfig.getGuiRegistry(ModConfig.class));
         INSTANCE = AutoConfig.getConfigHolder(ModConfig.class).getConfig();
     }
 
     @ConfigEntry.Gui.Tooltip
     @ConfigEntry.Gui.CollapsibleObject
     public MovementConfig movement = new MovementConfig();
+
```
MattSturgeon, line 28:
```suggestion
```

common/src/main/java/net/xolt/freecam/config/Behavior.java
```diff
@@ -0,0 +1,37 @@
...
+    public static final ModConfig CONFIG = ModConfig.INSTANCE;
+    public static final Variant BUILD_VARIANT = Variant.getInstance();
+
+    private Behavior() {}
+
+    public static boolean checkInitialCollision() {
+        return ModConfig.INSTANCE.collision.alwaysCheck || !(ModConfig.INSTANCE.collision.ignoreAll && BUILD_VARIANT.cheatsPermitted());
+    }
+
+    public static boolean allowFreezePlayer() {
```
MattSturgeon, line 22: Rearrange, place next to `freezePlayer(Entity)`

common/src/main/java/net/xolt/freecam/config/Behavior.java
```diff
@@ -0,0 +1,37 @@
...
+        return CONFIG.utility.allowInteract && (BUILD_VARIANT.cheatsPermitted() || CONFIG.utility.interactionMode.equals(PLAYER));
+    }
+
+    public static boolean disableInteract() {
+        return Freecam.isEnabled() && !Freecam.isPlayerControlEnabled() && !allowInteract();
+    }
+
+    public static boolean freezePlayer(Entity entity) {
+        return Freecam.isEnabled() && entity.equals(MC.player) && allowFreezePlayer();
+    }
```
MattSturgeon, line 36:
```suggestion
    public static boolean freezeEntity(Entity entity) {
        return Freecam.isEnabled() && entity.equals(MC.player) && allowFreezePlayer();
    }
```

variant/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+subprojects {
+    if (project.name != "api") {
+        architectury {
+            common(rootProject.enabled_platforms.split(','))
+        }
+
+        dependencies {
+            api project(":variant:api")
+        }
+    }
```
MattSturgeon, line 18: Add `fabric-loader` compileOnly dependency to `:api` to prevent missing class errors; something is referencing fabric's [`@Environment` annotation](https://maven.fabricmc.net/docs/fabric-loader-0.13.3/net/fabricmc/api/Environment.html).

metadata/build.gradle
```diff
@@ -0,0 +1,77 @@
...
+
+        archiveClassifier = classifier
+
+        // Include files under `src/platform`
+        from(layout.projectDirectory.dir("src").dir(platform))
+
+        inputs.properties(
+                mod_id: rootProject.name,
+                modrinth_name: project.modrinth_name,
+                version: project.mod_version,
```
MattSturgeon, line 37: Should the version shown in-game (and defined in mod metadata) be the `mod_version` or the _full_ version? IIRC it was previously the full version (including the mc version.
```suggestion
                version: project.version,
```

hashalite:
Hey, thanks again for putting in the effort to do this. The number of branches was really getting out of hand :)

When running from my IDE, everything appears to work flawlessly. However, the jars produced all crash with varying errors when launched externally. Here are the errors I received:

fabric:
https://pastebin.com/mGDs7Hp5

fabric-modrinth:
https://pastebin.com/YBPYA5AJ

forge:
https://pastebin.com/XKtYHHP1

forge-modrinth:
https://pastebin.com/zXcWENGn

Could I be doing something incorrectly? I am simply running `gradle :fabric:build` and `gradle :forge:build`

metadata/build.gradle
hashalite: Yes, I think mod_version is better. Not sure why I had it using the full version.

MattSturgeon:
Oops, knew I should've checked prod builds...

Looks like I'm not shadowing the dependencies correctly so some stuff is missing from the jar;

`Variant` is missing for normal builds and the freecam classes seem to be missing for modrinth builds.

I'll look into it when I get chance, but should be fairly simple 🤞

In the mean time any other questions or review points are welcome 😅

forge/build.gradle
```diff
@@ -31,35 +36,93 @@ dependencies {
...
+
+    // Add a dependency on metadata
+    def metadata = dependencies.project(path: ":metadata", configuration: "forge${variant == "normal" ? "" : variant.capitalize()}")
+    dependencies.add(set.implementationConfigurationName, metadata)
+    dependencies.add(shadowConfig.name, metadata)
+
+    // Add a dependency on variant's service provider
+    def spi = dependencies.project(path: ":variant:${variant}", configuration: "namedElements")
+    dependencies.add(set.implementationConfigurationName, spi)
+    dependencies.add(shadowConfig.name, dependencies.project(path: ":variant:${variant}", configuration: "transformProductionForge"))
```
MattSturgeon, line 63: Consider using `include` (JarJar/JarInJar) instead of shadow here?

forge/build.gradle
```diff
@@ -31,35 +36,93 @@ dependencies {
...
+    shadowJarTask.configure {
+        group = "shadow"
+        from sourceSets.main.output
+        from set.output
+        exclude "architectury.common.json"
 
-    configurations = [project.configurations.shadowCommon]
-    archiveClassifier = "dev-shadow"
-}
+        configurations = [
```
MattSturgeon, line 107: Also need:

```gradle
manifest {
    inheritFrom(tasks.jar.manifest)
}
```

MattSturgeon: Can probably drop `shadow` again if it inherites `jar`s manifest.

MattSturgeon:
@hashalite the latest force-push should fix everything. :crossed_fingers: 

- `shadowJarTask` needed to use both the main & variant source sets
- `shadowJarTask` needed to use the jar manifest that loom adds to the normal `jar` task.
- Needed to explicitly shadow `:variant:api` because it being a dependency of the implementation wasn't enough for it to get shadowed.

I also dropped the `Behavior` class for now. That refactor felt a little out-of-scope for this PR and also kinda half-baked.

hashalite:
Another thing: it seems like the access widener for renderShadows is not working in the production builds. It crashes with the following error when Freecam is toggled:

java.lang.IllegalAccessError: class net.minecraft.client.renderer.entity.EntityRenderer tried to access private field net.minecraft.client.renderer.entity.EntityRenderDispatcher.f_114368_

MattSturgeon:
Good catch.

Is that just forge or fabric too? Normal and Modrinth variants?

Fabric should just use the AccessWidener, but for forge it's supposed to be translated into Access transformer syntax.

hashalite:
It appears to happen with all versions except the regular forge build, strangely

hashalite:
One other nit-picky request: is there any way we could stop the empty `freecam--1.2.1.1+1.x.x` and `freecam--1.2.1.1+1.x.x-sources` jars from being generated in the root build directory when `:build` is run? They don't appear to be necessary (correct me if I'm wrong)

hashalite:
I have just migrated the project over to official mappings, hopefully the conflicts will not be too difficult to resolve

MattSturgeon:
> it seems like the access widener not working in the production builds. It crashes when Freecam is toggled
> It appears to happen with all versions except the regular forge build, strangely

That _is_ strange... I'll investigate shortly. May have to ask in the architectury discord again.

> is there any way we could stop the empty [root project] jars from being generated when :build is run?

Yes, was planning to do this actually.  `tasks.jar.enabled = false` should work, just need to check it won't break subprojects running their tasks too.

> I have just migrated the project over to official mappings, hopefully the conflicts will not be too difficult to resolve

Should just be able to do `./gradlew migrateMappings –mappings "net.mojang.minecraft:mappings:1.20.4"` on each commit.

Might make a PR adding a Parchment layer to the mappings, if you don't beat me to it. Would be nice to have the javadocs & parameter names.

Should just be add their maven & then:
```gradle
mappings loom.layered() {
    officialMojangMappings()
    parchment("org.parchmentmc.data:parchment-1.20.2:2023.12.10@zip")
}
```
Hopefully their 1.12.2 layer (latest) still works ok on 1.12.4 🤞

MattSturgeon:
> it seems like the access widener not working in the production builds. It crashes when Freecam is toggled
> It appears to happen with all versions except the regular forge build, strangely

Figured out the issue, but still unclear on the solution.

There's two things going on here.

### Fabric

On fabric, architectury is supposed to add the `accessWidener` entry in `architectury.common.json` to `fabric.mod.json`, however because we're building the mod metadata in `:metadata` now, this doesn't happen.

### Forge

On forge, Architectury is supposed to remove all AccessWideners from the build and use their content to generate a `META-INF/accesstransformer.cfg`. This is happening, but only for the default build variant. Architectury doesn't know about our modrinth build.

MattSturgeon:
This should be all fixed now. I've tested the modrinth variant in both forge & fabric. Toggling works normally.

I still need to rebase this on main & fix conflicts, however I'm running into issues with `migrateMappings`... Not sure if its a loom bug or an issue on my end: https://github.com/architectury/architectury-loom/issues/188

hashalite:
> I'm running into issues with `migrateMappings`... Not sure if its a loom bug or an issue on my end

I think it may be an issue on your end, I was able to run the same command yesterday without issue

MattSturgeon:
Strange. Just tried a second machine - same issue. Wiping cache doesn't help either.

Maybe OS specific? Might try in a VM if I get chance :thinking:

If  `migrateMappings` still works for you, you could push a commit  to this branch converting to 1.20.2 mappings. I _should_ be able to use that as a reference for a manual rebase.

hashalite:
> If `migrateMappings` still works for you, you could push a commit to this branch converting to 1.20.2 mappings

Done! Although the mixins still need to be fixed since they aren't migrated automatically

MattSturgeon:
> > If `migrateMappings` still works for you, you could push a commit to this branch converting to 1.20.2 mappings
> 
> Done!

Thanks. The rebase ended up being simpler than I was expecting. The mappings only really affected `ConfigExtensions` (`Text` -> `Component`).

1.20.4 forge dev runs are still broken, so its harder to test this PR now, but it was working fine before rebase so should be ok... Only seen one mention of a similar issue in the architectury discord, so other people may be unaffected :thinking: 

Tested prod builds again, everything _seems_ to be working.

hashalite:
> The rebase ended up being simpler than I was expecting.

Glad to hear it!

Will test further and get merged asap.

Thanks again, and merry xmas!

MattSturgeon:
> Thanks again, and merry xmas!

Merry Christmas to you too :christmas_tree:

PR #136, opened 2023-12-27
1.16.5: Compile for Java 8

MattSturgeon:
This fixes class version errors when attempting to launch the prod build.

There are still other errors this doesn't fix yet:

- [ ] Cloth Config is missing in production (i.e. JarJar isn't working)
- [ ] Dev-Env runs are broken (complaining about `'com.mojang.math.Transformation com.mojang.math.Transformation.func_227987_b_()'`).

The first issue should be fix-able using shadow.  EDIT: this will be a bit of a pain, because loom won't convert our AW to AT if cloth-config's AT is already in the jar... We could _really_ hack-fix that by including cloth's transforms in our AW and excluding their AT in the `shadowJar` task...

The second issue could be an arch-loom bug affecting older MC versions?

Incidentally, Java8's version of `SingleInstanceServiceLoader` _might_ be nicer than the version I originally wrote using streams? :thinking:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -84,7 +84,7 @@ allprojects {
     // see http://yodaconditions.net/blog/fix-for-java-file-encoding-problems-with-gradle.html
     tasks.withType(JavaCompile) {
         options.encoding = "UTF-8"
-        options.release = 17
+        options.release = 8
     }
 
     java {
--- a/common/src/main/java/net/xolt/freecam/config/ConfigExtensions.java
+++ b/common/src/main/java/net/xolt/freecam/config/ConfigExtensions.java
@@ -66,7 +66,7 @@ public class ConfigExtensions {
     @Contract(mutates = "param1")
     private static void applyVariantTooltip(List<AbstractConfigListEntry> guis, List<VariantTooltip> tooltipVariants, String i18n) {
         String variant = BuildVariant.getInstance().name();
-        List<String> variants = List.of(variant, "all");
+        List<String> variants = Arrays.asList(variant, "all");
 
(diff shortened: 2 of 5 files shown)


MattSturgeon:
Got shadow kinda working locally, needed to add cloth-config's access widener to our repo and exclude all accesstransformers from the shadow jar.

I haven't pushed here as I'm hitting a blocker merging the `mods.toml` files. You can find my in-progress work [here](https://github.com/MattSturgeon/Freecam/commits/1.16-fixes-w/shadow/).

Unless I'm being dense, the only way to merge/remove/whatever the shadowed `mods.toml` files is telling shadow to use a `Transformer` ([small example](https://github.com/johnrengelman/shadow/blob/9c5182d2d9c5f7141e4d2a525ec94d6111283cd9/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/AppendingTransformer.groovy)).

The transformer would have to read the file's lines, determine if the global properties are ours (the ones before `[[mods]]`), then read the rest of the file too. When writing to output it would write _our_ global properties (above `[[mods]]`) and _all_ other properties ( `[[mods]]` and below) to the output zip entry.

This shouldn't be too difficult. The question is "is it worth it" when the alternative is telling the handful of 1.16 forge users to install cloth-config as an external dependency...

hashalite:
Will go ahead and merge this as I think shadowing Cloth should be its own PR if we decide to do it. 

> In the mean time, we should test if having cloth-config installed as a mod is enough to get this build running.

I tested with cloth installed separately, and everything appears to function properly

PR #137, opened 2023-12-28
Store whitelist conditions as predicate functions

MattSturgeon:
This small refactor enables the collision whitelist to be more flexible.

The predicate builder now supports:  
- Matching on block class (existing behavior).
- Matching on block ID string.
- Regex matching on block ID string.
- Matching using an arbitrary predicate function.

This PR could be expanded or followed up with:  
- More explicit whitelists using block ID matching.
- User-configurable whitelists/blacklists edited vla a cloth-config.
  - Autoconfig creates a [`StringListListEntry`](https://github.com/shedaniel/cloth-config/blob/461ae8e56fa02f0422e2b8a3b1f38d743e5fba7e/common/src/main/java/me/shedaniel/clothconfig2/gui/entries/StringListListEntry.java#L35) for `List<String>` and `String[]` config fields.
  - I've [played around a little](https://github.com/MattSturgeon/Freecam/blob/d83662bd33f4d3a5a685fc54b92199b1923654d4/common/src/main/java/net/xolt/freecam/config/gui/ValidateRegexImpl.java#L37) with adding an "error supplier" to validate whitelist regex patterns.

### DIFF
--- a/common/src/main/java/net/xolt/freecam/config/CollisionWhitelist.java
+++ b/common/src/main/java/net/xolt/freecam/config/CollisionWhitelist.java
@@ -1,36 +1,73 @@
 package net.xolt.freecam.config;
 
-import net.minecraft.world.level.block.BarrierBlock;
-import net.minecraft.world.level.block.Block;
-import net.minecraft.world.level.block.DoorBlock;
-import net.minecraft.world.level.block.FenceGateBlock;
-import net.minecraft.world.level.block.IronBarsBlock;
-import net.minecraft.world.level.block.TransparentBlock;
-import net.minecraft.world.level.block.TrapDoorBlock;
+import net.minecraft.core.registries.BuiltInRegistries;
+import net.minecraft.world.level.block.*;
+
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
-import java.util.List;
+import java.util.function.Predicate;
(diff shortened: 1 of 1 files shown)


common/src/main/java/net/xolt/freecam/config/CollisionWhitelist.java
```diff
@@ -1,36 +1,65 @@
...
+        }
+
+        public final Builder matching(String... ids) {
+            return matching(block -> Arrays.asList(ids).contains(block.getDescriptionId()));
+        }
+
+        public final Builder matching(Pattern... patterns) {
+            return matching(block -> Arrays.stream(patterns)
+                    .map(pattern -> pattern.matcher(block.getDescriptionId()))
+                    .anyMatch(Matcher::hasMatch));
```
MattSturgeon, line 48: Should use `Matcher::find`

MattSturgeon: Should also use `BuiltInRegistries.BLOCK.getKey(block).toString()` since `getDescriptionId` looks like `block.minecraft.air`.

MattSturgeon: Done

hashalite:
Looks good, thank you!

PR #138, opened 2023-12-28
Add Parchment mappings layer

MattSturgeon:
This PR adds parchment mappings.

Mojang's mappings cover **class**, **method**, & **field** names. Parchment builds on this by providing mappings for **parameter** names and some **javadocs**, though not nearly as much javadoc as yarn currently has.

I slightly over-complicated things by making the parchment _minecraft_ version optional:

- If `parchment_version` contains a hyphen, the first part is the minecraft version, appended to the artifact name.
- If there's no hyphen, then the project `minecraft_version` is used instead.

This means when parchment has a release exactly matching the minecraft version we're using, we don't need to specify it twice.

For example, we are using 1.20.4, however Parchment has a stable build for 1.20.2 and unstable builds for 1.20.3. Therefore we currently need to specify two different mc versions.

If you don't think the complexity is worth it, I can remove the complexity and we'll always specify parchment's mc version explicitly.

We may also want to be able to _disable_ parchment. We could do this by adding a check for `project.parchment_version == "off"` (or similar). Or we could comment the parchment line on an ad-hoc basis.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,8 +47,22 @@ subprojects {
     }
 
     dependencies {
+        def parchmentAppendix
+        def parchmentVersion
+        def parts = rootProject.parchment_version.split('-')
+        if (parts.size() == 2) {
+            parchmentAppendix = parts[0]
+            parchmentVersion = parts[1]
+        } else {
+            parchmentAppendix = rootProject.minecraft_version
+            parchmentVersion = parts[0]
+        }
+
         minecraft "com.mojang:minecraft:${rootProject.minecraft_version}"
-        mappings loom.officialMojangMappings()
+        mappings loom.layered() {
+            officialMojangMappings()
+            parchment("org.parchmentmc.data:parchment-${parchmentAppendix}:${parchmentVersion}@zip")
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -47,8 +47,22 @@ subprojects {
...
+        def parchmentAppendix
+        def parchmentVersion
+        def parts = rootProject.parchment_version.split('-')
+        if (parts.size() == 2) {
+            parchmentAppendix = parts[0]
+            parchmentVersion = parts[1]
+        } else {
+            parchmentAppendix = rootProject.minecraft_version
+            parchmentVersion = parts[0]
+        }
```
MattSturgeon, line 59: If you don't care about having to specify parchment's mc version, we could simplify this:

```suggestion
        def parchmentAppendix = rootProject.parchment_version.split('-')[0]
        def parchmentVersion = rootProject.parchment_version.split('-')[1]
```

hashalite: Specifying the mc version is fine, I don't think the added complexity is worth it

PR #139, opened 2023-12-28
Migrate forge -> neoforge

### LINKED ISSUE #133: Neoforge
The current consensus in forge modding is to migrate away from Lex's Forge and instead use Neoforge, which is maintained by the majority of the original forge team.

Many projects are outright switching and dropping Forge support. Some with 1.20.4, some already have done in a previous release, others plan to do so in a future release.

Some projects are maintaining both Forge & Neoforge. In the short term, this is probably best for users. However, long term it'd be better if the most popular mods were Neoforge exclusive, as this would hopefully reduce fragmentation in the ecosystem.

Supporting both isn't a big deal; we would simply add a `:neoforge` gradle project. Most of the code would be identical to `:forge` however.

It is possible to have a shared `:forge-like` where shared code/mixins can live, however it is a bit cumbersome and requires custom mappings patches due to some Forge/Neoforge differences.

So long as the forge-like code is small, I think it's simpler to just duplicate it across both Forge projects. Especially if you plan on dropping one of those projects later.

See the [Architectury migration guide](https://docs.architectury.dev/api/migration/neoforge#migrate_your_architectury_project_to_neoforge) for more detail.

### LINKED ISSUE #135: Forge dev-env runs broken in 1.20.4
Forge (development) runs now fail with a `ResolutionException`; essentially two (identical) modules are providing the same packages, so forge panics.

Example error (from another project with the same issue as I'm not currently at a PC):

```
<==Exception in thread "main" java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
        at net.minecraftforge.bootstrap.Bootstrap.start(Bootstrap.java:48)
        at net.minecraftforge.bootstrap.ForgeBootstrap.main(ForgeBootstrap.java:18)
        at net.fabricmc.devlaunchinjector.Main.main(Main.java:86)
        at dev.architectury.transformer.TransformerRuntime.main(TransformerRuntime.java:219)
Caused by: java.lang.module.ResolutionException: Modules generated_81acd6d and animal.feeding.trough._1._20._4.dev export package slexom.animal_feeding_trough.platform.common.block.entity to module animal_feeding_trough
```

See [this thread](https://discord.com/channels/792699517631594506/1183085737940361356/1183085737940361356) on the Architectury discord.

One proposed workaround (others reported success, I've not yet tested) is removing `runtimeClasspath.extendsFrom common`.

_I guess_ both `runtimeClasspath` **and** `developmentForge` extend from `common`, so the classes are loaded twice at runtime.  
If so, I'm surprised more people aren't reporting this issue.

Opening this to keep track of the issue until I have time to put together a PR.

MattSturgeon:
This PR migrates the forge project to neoforge.

I've noticed a couple issues so far _(both fixed now)_:

- [x] Dev runs _still_ don't work: Neoforge complains "`Failed to find system mod: minecraft`".
- [x] Show player is broken _again_. Player isn't rendered, regardless of "show player" setting.

Fixes #133 
Fixes #135


gradle.properties
```diff
@@ -26,10 +26,10 @@ fabric_api_version=0.91.2+1.20.4
 fabric_loader_req=>=0.12.11
 fabric_mc_req=~1.20.4
 
-forge_version=1.20.4-49.0.9
+forge_version=20.4.60-beta
 forge_mc_req=[1.20.4,)
-forge_loader_req=[49,)
-forge_req=[49,)
+forge_loader_req=[1,)
+forge_req=[20,)
```
MattSturgeon, line 32: Should these be renamed to `neoforge_*`?

neoforge/build.gradle
```diff
@@ -10,43 +9,25 @@ def variants = rootProject.build_variants.split(',')
...
-        mixinConfig "freecam-forge.mixins.json"
-    }
-}
-
-afterEvaluate {
-    // FIXME: arch-loom only does this for tasks.remapJarTask,
-    //      meaning our variant tasks don't get AccessTransformers.
-    tasks.withType(RemapJarTask).each {
-        Aw2At.setup(project, it)
-    }
```
MattSturgeon, line 31: Glad all this is gone.

The `afterEvaluate` block could be removed on non-neoforge branches too, by adding `atAccessWideners.add loom.accessWidenerPath.get().asFile.name` to the `remapJarTask` config.

MattSturgeon:
I was able to get past the dev-run issue by removing the deprecated 'mandatory' field (used `type` instead).

I also ran this gradle command for good measure: 

```shell
./gradlew -Dloom.refresh=true tasks
```

Devlogin turned out to be an issue too, I've disabled it for now. See https://github.com/PlanetTeamSpeakk/DevLogin/issues/16.

MattSturgeon:
Reverting the LexForge 'Show Player' fix was enough. I guess Neoforge doesn't have that quirk.

gradle.properties
hashalite: Yes, I think that would help to avoid confusion as older versions will continue using LexForge

MattSturgeon: Done

hashalite:
Much appreciated!

PR #140, opened 2023-12-28
Mod platform publishing task

### LINKED ISSUE #134: Gradle publishing
[This plugin](https://github.com/firstdarkdev/modpublisher) aims to simplify publishing artifacts to GitHub releases, curseforge & modrinth.

Once set up, publishing a release would be done by running a gradle publish task. The secret keys for each platform would need to be set as environment variables (or similar).

The plugin author has [added](https://github.com/firstdarkdev/modpublisher/issues/3) the ability to override the artifact for a specific platform, meaning we can upload the restricted build to modrinth.

There are alternative plugins that do similar things. E.g. shedanial maintains [`unified-publishing`](https://github.com/shedaniel/unified-publishing), ~~which I haven't looked into yet~~. _EDIT: looks equally straightforward, perhaps a bit more fleshed out._

You can also use the various APIs manually, however a plugin centralised a lot of shared configuration and streamlines the process.

MattSturgeon:
# Mod publishing

This PR adds the [modpublisher plugin](https://github.com/firstdarkdev/modpublisher) which simplifies the process of publishing to [CurseForge](https://www.curseforge.com/) & [Modrinth](https://modrinth.com/).

For an overview, see [their readme](https://github.com/firstdarkdev/modpublisher/blob/2.0/readme.md) or their [full documentation](https://modpublisher.fdd-docs.com).

## Usage

After creating a release (setting version, release type, patching changelog, committing), you will use the `publish` task.

`publish` is Gradle's "umbrella" task, which depends on all publishing tasks. If you only want to publish to modding platforms, you can use `publishMod` instead, or even `publishCurseforge` & `publishModrinth` to publish only to Curseforge and Modrinth respectively.

Note there is also a `publishGitHub` task provided by modpublisher, however this will do nothing because we haven't configured modpublisher for GitHub Releases (see below). We can't disable the task (which would hide it) because `publishMod` depends on it.

## GitHub Releases

Publishing to GitHub is not covered by this PR. That may be better achieved as part of a CI/CD setup using GitHub Actions/Workflows.

[modpublisher](https://github.com/firstdarkdev/modpublisher) has built-in support for publishing GitHub releases, however for our use case the "version" would need to be different and we'll probably need more control...

## Secrets

API tokens/keys must be set to actually do any publishing. These can either be gradle properties (`curseforge_token` & `modrinth_token`) or environment variables (`CURSEFORGE_TOKEN` & `MODRINTH_TOKEN`).

Personally, I'd probably set these as "user" gradle properties by creating a `gradle.properties` file in the [`GRADLE_USER_HOME`](https://docs.gradle.org/current/userguide/directory_layout.html#dir:gradle_user_home) directory (`~/.gradle` or `C:\Users\<USERNAME>\.gradle`). Setting up `.env` is another option.  
Definitely **do not** put them in the project gradle properties, as these get committed to git :joy:!

We'll also want these saved in the **repo secrets** if we ever setup a CI job to automate publishing.

## Published Jars

The appropriate jars are published to each platform; all platforms use the normal build variant (`remapJar`) except for Modrinth, which uses the output from `remapModrinthJar`.

## Changelog

I haven't added anything to the changelog as this PR only affects internal (dev-facing) stuff, which is already mentioned in the changelog.

The published release will contain the changelog entry relating to the current `mod_version`.

The changelog section header is omitted, because that _should_ be self-evident from the release the changelog is attached to, however this is configurable and could be included if desired.

## Alternative plugins

We initially explored using shedanial's [unified-publishing](https://github.com/shedaniel/unified-publishing), however we ran into an [issue](https://github.com/shedaniel/unified-publishing/issues/8) setting "embedded" related mods for the Modrinth platform. Seeing as everything we needed worked on  [modpublisher](https://github.com/firstdarkdev/modpublisher) & that project appears to be more actively maintained, we decided to use that instead.

### Original PR description
<details><summary>Hidden for brevity</summary>

This PR implements shedanial's [unified-publishing](https://github.com/shedaniel/unified-publishing) plugin to streamline the curseforge/modrinth publishing process.

Publishing to GitHub is not covered by this PR. That may be better achieved as part of a CI/CD setup using GitHub Actions/Workflows.  
That said, [modpublisher](https://github.com/firstdarkdev/modpublisher) has built-in support for publishing GitHub releases as part of a unified publish task, so we could consider using that instead.

API tokens/keys must be set to actually do any publishing. These can either be gradle properties (`curseforge_token` & `modrinth_token`) or environment variables (`CURSEFORGE_TOKEN` & `MODRINTH_TOKEN`).

Personally, I'd probably set these as "user" gradle properties by creating a `gradle.properties` file in the [`GRADLE_USER_HOME`](https://docs.gradle.org/current/userguide/directory_layout.html#dir:gradle_user_home) directory (`~/.gradle` or `C:\Users\<USERNAME>\.gradle`). Setting up `.env` is another option.  
Definitely **do not** put them in the project gradle properties, as these get committed to git :joy:!

Run using `gradlew publishUnified`.

**I haven't tested this**, since I don't have access to your tokens. It's possible I've misinterpreted what is meant by "modrinth id"; currently it's set to `"freecam"`.

When testing, I'd recommend changing the `releaseType` to alpha and changing `mod_version` to something like `"TEST"`.

Come to think of it, `releaseType` should probably be set as in `gradle.properties`... Any other feedback, questions, etc welcome.

</details> 

Fixes #134

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,37 +52,6 @@ changelog {
     headerParserRegex = ~/(\d+(?:\.\d+)+)/
 }
 
-subprojects {
-    apply plugin: "dev.architectury.loom"
-
-    loom {
-        accessWidenerPath = project(":common").file "src/main/resources/freecam.accesswidener"
-
-        runs {
-            client {
-                // Pretty run-config name
-                // Note: we can't disable the (:path) suffix
-                // https://github.com/architectury/architectury-loom/blob/5b3e7c72b665f7eb38c23ceb677027acdc867398/src/main/java/net/fabricmc/loom/configuration/ide/RunConfig.java#L117
-                configName "Run"
-            }
-            remove server // We only need client runs
-        }
-
-        silentMojangMappingsLicense()
-    }
-
-    dependencies {
-        def parchmentAppendix = rootProject.parchment_version.split('-')[0]
-        def parchmentVersion = rootProject.parchment_version.split('-')[1]
-
-        minecraft "com.mojang:minecraft:${rootProject.minecraft_version}"
-        mappings loom.layered() {
-            officialMojangMappings()
-            parchment("org.parchmentmc.data:parchment-${parchmentAppendix}:${parchmentVersion}@zip")
-        }
-    }
-}
-
 allprojects {
     apply plugin: "java"
     apply plugin: "architectury-plugin"
@@ -125,5 +94,36 @@ allprojects {
 
     jar {
         from "LICENSE"
+    }
+}
+
+subprojects {
+    apply plugin: "dev.architectury.loom"
+
+    loom {
+        accessWidenerPath = project(":common").file "src/main/resources/freecam.accesswidener"
+
+        runs {
+            client {
+                // Pretty run-config name
+                // Note: we can't disable the (:path) suffix
+                // https://github.com/architectury/architectury-loom/blob/5b3e7c72b665f7eb38c23ceb677027acdc867398/src/main/java/net/fabricmc/loom/configuration/ide/RunConfig.java#L117
+                configName "Run"
+            }
+            remove server // We only need client runs
+        }
+
+        silentMojangMappingsLicense()
+    }
+
+    dependencies {
+        def parchmentAppendix = rootProject.parchment_version.split('-')[0]
+        def parchmentVersion = rootProject.parchment_version.split('-')[1]
+
+        minecraft "com.mojang:minecraft:${rootProject.minecraft_version}"
+        mappings loom.layered() {
+            officialMojangMappings()
+            parchment("org.parchmentmc.data:parchment-${parchmentAppendix}:${parchmentVersion}@zip")
+        }
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,5 @@
 import net.fabricmc.loom.task.RemapJarTask
+import org.jetbrains.changelog.Changelog
 
 plugins {
     id "architectury-plugin" version "3.4-SNAPSHOT"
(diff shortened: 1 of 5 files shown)


MattSturgeon:
@hashalite the next release could be a good opportunity to test this out? You could merge this locally and try using it without actually merging it on GitHub.

I'm happy to help worth through any teething issues with you on Discord, since I can't really test this myself without setting up Curseforge/Modrinth pages.

build.gradle
```diff
@@ -51,6 +54,43 @@ subprojects {
...
+                    }
+                }
+
+                curseforge {
+                    id = "557076"
+                    token = findProperty("curseforge_token") ?: System.getenv("CURSEFORGE_TOKEN") ?: ""
+                }
+
+                modrinth {
+                    id = "freecam"
```
hashalite, line 87: ID should be `XeEZ3fK2`

hashalite (changes requested):

MattSturgeon:
I've ported over to [modpublisher](https://github.com/firstdarkdev/modpublisher) after our conversation earlier.

Once this is tested/refined further I'll squash the commits, but I've kept it separate for now in-case we decide to revert.

The issue with the fabric project not correctly configuring the requirements might be resolved by putting it in an `afterEvaluate` block (if it isn't already). Otherwise we can fallback to moving them to the root gradle script under an `if (project.name == "fabric")` block.

hashalite (approved):

MattSturgeon:
I've polished a bit further and implemented the ability to list additional supported mc versions using `supported_mc_versions` etc. `minecraft_version` is always marked as supported.

I was able to do some limited testing by setting a dummy api key & `debug=true`, everything appeared to work correctly.

The [issue we were experiencing](https://github.com/firstdarkdev/modpublisher/issues/6) with modrinth dependencies _should_ be resolved, but I can't test that.

The issue with `project.version` not being defined is now resolved.

MattSturgeon:
Bumped to modpublisher 2.0.5, so the [modrinth slug->id issue](https://github.com/firstdarkdev/modpublisher/issues/6) should be fully resolved now.

Worked correctly in my tests:

```json
"dependencies": [
    {
      "projectId": "P7dR8mSH",
      "dependencyType": "required"
    },
    {
      "projectId": "mOgUt4GM",
      "dependencyType": "optional"
    },
    {
      "projectId": "9s6osm5g",
      "dependencyType": "embedded"
    }
  ],
```

hashalite (approved):
Everything looks good, thank you!

PR #141, opened 2023-12-28
Cleanup aw->at config

MattSturgeon:
While working on #139, I discovered `RemapJarTask.atAccessWideners`, which lets us avoid the messy accesstransformer transform config.

I've made this PR on the 1.19 branch because I assume #139 will make it redundant on the main branch. The earliest version supported by Neoforge is 1.20.2.


hashalite:
Thanks!

PR #143, opened 2023-12-30
Allow setting keybinds from Freecam's config

MattSturgeon:
## Goal

Allow users to configure everything from a single menu, if they prefer.

Keybinds are still present in the vanilla menu. This doesn't change how binds are stored/created/etc.

## Screenshot

image

## Implementation

This PR adds a new `AddModBindings` annotation, implemented by `GuiTransformer`s that add a `KeyCodeEntry`, for each `ModBindings` value, to the annotated field.

This is accompanied by some minor refactoring/renaming.

## Movement

IMO, everything in **Movement** is somewhat _controls_ related. Perhaps the two categories should be combined?

Perhaps **Key binds** and **Movement** could be subcategories of **Controls**?

If **Movement** is moved or renamed, it'd be a good opportunity to scale up & convert the movement speeds to `int`, so they can be sliders instead of text fields. Otherwise users would keep their scaled-down values when upgrading.

### DIFF
--- a/common/src/main/java/net/xolt/freecam/config/ModConfig.java
+++ b/common/src/main/java/net/xolt/freecam/config/ModConfig.java
@@ -9,6 +9,7 @@ import me.shedaniel.autoconfig.serializer.JanksonConfigSerializer;
 import me.shedaniel.clothconfig2.gui.entries.SelectionListEntry;
 import net.xolt.freecam.config.gui.AutoConfigExtensions;
 import net.xolt.freecam.config.gui.BoundedContinuous;
+import net.xolt.freecam.config.gui.ModBindingsConfig;
 import net.xolt.freecam.config.gui.VariantTooltip;
 import net.xolt.freecam.variant.api.BuildVariant;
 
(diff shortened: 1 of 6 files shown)


MattSturgeon:
Rebased & added changelog

common/src/main/java/net/xolt/freecam/config/ModConfig.java
```diff
@@ -24,6 +25,15 @@ public static void init() {
         INSTANCE = AutoConfig.getConfigHolder(ModConfig.class).getConfig();
     }
 
+    @ConfigEntry.Gui.Tooltip
+    @ConfigEntry.Gui.CollapsibleObject
+    public ControlsConfig controls = new ControlsConfig();
+    public static class ControlsConfig {
+        @ModBindingsConfig
+        @ConfigEntry.Gui.PrefixText
```
MattSturgeon, line 33: Not sure what this `PrefixText` is doing, can't see a i18n key for it. I guess I intended to have some preamble but never wrote it?

hashalite (approved):

PR #146, opened 2023-12-31
Server whitelist/blacklist

### LINKED ISSUE #117: Feature request | Server whitelist
There are many servers which forbid even the Modrinth version and also many servers also have in their rules that the single join just with the mod installed without using it is already prohibited. I think it's pretty tedious to just remove the hotkey every time you want to play on a minigames server for example. It also could happen that you accidentally hit your freecam hotkey in a live stream for example. There should be a whitelist option that can be toggled. If enabled, you can add IPs/domains that will enable the mod's functionality on added servers.

MattSturgeon:
## Overview

Includes the same config-extension refactoring commit from #143.

Allows the user to setup a whitelist or blacklist in their Freecam config.

If enabled, this will prevent freecam from activating on the restricted servers.

If notifications are enabled, `"Freecam is restricted on the current server (server_ip_here)."` will be shown when attempting to toggle freecam.

Not convinced my naming/wording (or my config layout) is optimal, so I'm open to feedback & suggestions.

I noticed when testing that cloth-config's `StringListListEntry` is kinda janky; adding a list-entry doesn't focus the new text field; the text field is invisible, so it's not obvious you can click on it. We may be able to work around this with some mixins and/or autoconfig extensions.

### Testing logged in

To test out using DevLogin, you'll want to uncomment the dependency and add this to line 74 of `fabric/build.gradle`:

```gradle
loom.runs.create("${variant}LoggedIn") {
    inherit run
    configName "Run logged in"
    programArg "--msa"
}
```

Then run `runNormalLoggedIn` or `runModrinthLoggedIn`. The `runLoggedIn` task created in the root `build.gradle` won't work (I'll fix that at some point, will probably wait for devlogin to finish their neoforge support).

Alternatively, use DevAuth by either dropping it into `mods` and setting `defaultEnabled` or `devauth.enabled` as required, or using the devauth PR.

Fixes #117


common/src/main/resources/assets/freecam/lang/en_us.json
```diff
@@ -66,6 +67,16 @@
   "text.autoconfig.freecam.option.utility.interactionMode.@Tooltip": "The source of block/entity interactions.",
   "text.autoconfig.freecam.option.utility.interactionMode.camera": "Camera",
   "text.autoconfig.freecam.option.utility.interactionMode.player": "Player",
+  "text.autoconfig.freecam.option.servers": "Server Options",
```
MattSturgeon, line 70: Perhaps "Multiplayer Options" would be better?

common/src/main/resources/assets/freecam/lang/en_us.json
```diff
@@ -66,6 +67,16 @@
...
+  "text.autoconfig.freecam.option.servers.@Tooltip": "Options relating to multiplayer servers.",
+  "text.autoconfig.freecam.option.servers.mode": "Restriction mode",
+  "text.autoconfig.freecam.option.servers.mode.@Tooltip[0]": "Restrict Freecam to certain servers",
+  "text.autoconfig.freecam.option.servers.mode.@Tooltip[1]": "(using either a whitelist or blacklist).",
```
MattSturgeon, line 74: This tooltip feels janky... Ideas?

common/src/main/java/net/xolt/freecam/Freecam.java
```diff
@@ -317,4 +332,16 @@ public static boolean isEnabled() {
...
+
+    public static boolean notAllowedInWorld() {
+        if (MC.isSingleplayer()) {
+            return false;
+        }
+
+        return switch (ModConfig.INSTANCE.servers.mode) {
+            case NONE -> false;
+            case WHITELIST -> !ModConfig.INSTANCE.servers.whitelist.contains(MC.getCurrentServer().ip);
+            case BLACKLIST -> ModConfig.INSTANCE.servers.blacklist.contains(MC.getCurrentServer().ip);
```
MattSturgeon, line 344: Should probably make this case-insensitive

MattSturgeon: Done

common/src/main/java/net/xolt/freecam/config/ModConfig.java
```diff
@@ -95,6 +101,17 @@ public static class UtilityConfig {
...
 
+    @ConfigEntry.Gui.Tooltip
+    @ConfigEntry.Gui.CollapsibleObject
+    public ServerConfig servers = new ServerConfig();
+    public static class ServerConfig {
+        @ConfigEntry.Gui.Tooltip(count = 2)
+        @ConfigEntry.Gui.EnumHandler(option = EnumDisplayOption.BUTTON)
+        public ServerRestriction mode = ServerRestriction.NONE;
+        public List<String> whitelist = Collections.emptyList();
+        public List<String> blacklist = Collections.emptyList();
```
MattSturgeon, line 112: These might benefit from some `PrefixText` or a `Tooltip`. Perhaps something along the lines of

> Add server IPs to this list using the button marked ⊕.
> Freecam will only be usable on servers whose IP exactly matches something in this list (case-insensitive).

common/src/main/java/net/xolt/freecam/config/gui/ServerRestrictionDependencies.java
```diff
@@ -0,0 +1,52 @@
...
+                },
+                field -> field.getDeclaringClass().equals(ModConfig.ServerConfig.class) && field.getName().equals("mode")
+        );
+
+        // Whitelist dependency
+        registry.registerPredicateTransformer(
+                (guis, i18n, field, config, defaults, guiProvider) -> {
+                    guis.stream()
+                            .filter(StringListListEntry.class::isInstance)
+                            .map(StringListListEntry.class::cast)
```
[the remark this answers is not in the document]
MattSturgeon: Done

common/src/main/java/net/xolt/freecam/config/ModConfig.java
MattSturgeon, line 112: These should be initialized with a mutable list instead of `Collections.emptyList()`.

Turns out autoconfig tries to mutate the existing list on load, and fails silently.

hashalite (approved):

hashalite (approved):

hashalite (approved):

PR #147, opened 2024-01-02
DevLogin -> DevAuth

MattSturgeon:
DevAuth has added neoforge support in their latest release, so I've ported from DevLogin.

`config.toml` is stored in `~/.config/devauth` or `C:\Users\<user>\.devauth`. You can have multiple accounts configured there.

You can also set `defaultEnabled = true` and dump the [devauth jar](https://github.com/DJtheRedstoner/DevAuth/releases) in `run/mods/` when debugging other projects, that way the project doesn't have to be configured to support devauth.

See the [configuration section](https://github.com/DJtheRedstoner/DevAuth#configuration) of their README for more detail.

You can copy IDE run configs and add `-Ddevauth.account=someaccount` if you have multiple accounts configured in your devauth config, for example.

Note: I've put this together on my laptop, which has an opengl issue meaning I can't test/run the game. Therefore **this PR is currently untested**.

### DIFF
--- a/neoforge/build.gradle
+++ b/neoforge/build.gradle
@@ -76,6 +76,13 @@ variants.each { variant ->
         dependency spi
     }
 
+    loom.runs.create("${variant == "normal" ? "" : variant}LoggedInClient") {
+        inherit run
+        forgeTemplate "client"
+        configName += " (logged in)"
+        property "devauth.enabled", "true"
+    }
+
     // Configure/create jar tasks
     def shadowJarTask
     def remapJarTask
--- a/neoforge/build.gradle
+++ b/neoforge/build.gradle
@@ -67,6 +67,7 @@ variants.each { variant ->
         }
     }
     run.source set
+    run.property "devauth.enabled", "false"
 
     // Tell forge all this is part of one mod
     run.mods.register(rootProject.name) {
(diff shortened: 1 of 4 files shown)


neoforge/build.gradle
```diff
@@ -23,6 +23,8 @@ configurations {
 dependencies {
     neoForge "net.neoforged:neoforge:${rootProject.neoforge_version}"
 
+    modRuntimeOnly "me.djtheredstoner:DevAuth-neoforge:${rootProject.devauth_version}"
```
MattSturgeon, line 26: backporting note: use `DevAuth-forge-latest` not `DevAuth-forge`. This is because DevAuth supports modern forge as well as pre-1.13 forge (`DevAuth-forge-legacy`).

MattSturgeon:
If you prefer, we can drop the "logged in" run configs, and instead rely on setting up `defaultEnabled` manually based on preference.

We could even drop baked-in devauth support entirely, since you can just drop the jar into `run/mods`...

hashalite:
> In fact, we could even drop baked-in devauth support entirely, since you can just drop the jar into `run/mods`...

Honestly, I think this would be the better option. Less maintenance going forward, and keeps the run configs tidy.

MattSturgeon:
> > In fact, we could even drop baked-in devauth support entirely
>
> Honestly, I think this would be the better option.

Ok, I'll do that.

PR #150, opened 2024-01-04
Drop DevLogin

MattSturgeon:
This PR started off as just dropping DevLogin; as per [this comment](https://github.com/hashalite/Freecam/pull/147#issuecomment-1876462140), however I figured it'd be nice to document the ability to run logged in via DevAuth... `CONTRIBUTING.md` snowballed from there, although I've left it unfinished as it isn't the focus of this PR.

TBH most of the content should probably be replaced with links to external documentation. Having a wall of text covering stuff that most devs probably already know is only going to mask the useful info such as multi-platform & multi-variant info.

I'll let you decide if you'd rather merge with/without the documentation. You could just cherry-pick 98d13ef7b82a1accdb0d0911b41d1ea196f6ed1a into main.

If you like the idea of a CONTRIBUTING doc, but not this iteration, we can always look at it again in another PR. Alternatively, this could be the starting point and another PR could improve on it.

EDIT: removed the documentation for now.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,14 +32,6 @@ subprojects {
                 // https://github.com/architectury/architectury-loom/blob/5b3e7c72b665f7eb38c23ceb677027acdc867398/src/main/java/net/fabricmc/loom/configuration/ide/RunConfig.java#L117
                 configName "Run"
             }
-            loggedIn {
-                inherit client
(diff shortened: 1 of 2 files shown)


hashalite (approved):
Thank you!

PR #152, opened 2024-01-06
Improve & generalise keybind code

MattSturgeon:
Building on #126's `ModBindings` refactor & toggle key handling, this PR:

- Generalizes the "combo key" logic in `FreecamComboKeyMapping`
- Moves control over keybind handlers to the keybinds themselves
- Introduces a builder to make defining different types of keybinds more declarative.
- Reduce the amount of "key handling" in `Freecam` to `activateTripodHandler()` & `resetTripodHandler()`,

I also removed `ModBinding`'s "wrapper" methods;  
They only existed to make using the enum values _prettier_, that is now unnecessary because the keybinds to define their own logic & don't care about the enum.  
The one exception to this is in `MinecraftMixin`, however a couple `.get()`s is worth the reduced complexity/maintenance IMO.  
If you disagree, then 00a5b7392f4bd826d50b7ab6c59495a58f56c676 can be dropped or reverted easily.

This PR does not affect the existing behavior in any way.

### DIFF
--- a/common/src/main/java/net/xolt/freecam/Freecam.java
+++ b/common/src/main/java/net/xolt/freecam/Freecam.java
@@ -1,14 +1,15 @@
 package net.xolt.freecam;
 
-import me.shedaniel.autoconfig.AutoConfig;
 import net.minecraft.client.CameraType;
 import net.minecraft.client.KeyMapping;
 import net.minecraft.client.Minecraft;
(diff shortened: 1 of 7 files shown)


hashalite (approved):
Thanks, this is much better. Handling the keybinds directly in the tick was pretty sloppy. This makes it much easier to see what exactly each keybind does.

PR #153, opened 2024-01-06
Initial unit tests

### LINKED ISSUE #151: Unit testing
Minecraft is a bit of a pain for unit tests, you can't easily pluck out a single class without first initializing the rest of the game.

That said, starting to add unit tests will make it easier to make changes more confidently, and make it easier to update to new versions/maintain multiple versions.

Long term, we'll end up improving the code base by refactoring classes to be more "testable", but initially we can start by just getting JUnit setup and testing the low hanging fruit like `FreecamPosition`.

I'm not sure what options are available to help setup tests and mock Minecraft classes. I've seen [this](https://github.com/BuiltBrokenModding/MinecraftJUnit), but it only targets older versions of the game.

MattSturgeon:
Beginning work on unit tests, opening a **draft PR** for early feedback.

The key breakthrough was calling `Bootstrap.bootStrap()` before running tests; without bootstrapping minecraft, registries (etc) are not initialized and **essentially no MC classes** can be constructed. This apparently affects 1.17 & newer.  
My solution is based on [this GPL3 code](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/d75ed36f21c8b51e58ddff5a22299e45c19976be/src/test/java/appeng/util/BootstrapMinecraftExtension.java), _see comments below_.

The goal of this PR **is not** to provide full test coverage, nor is this PR attempting to add integration tests with minecraft dependencies. That can all come in the future.

Instead this PR aims to enable basic unit testing functionality and provide one or two examples as a starting point.

Integration tests will likely require a `test` mod/run configured in loom, since the game will actually have to run to test integration. We might be able to do something with Mojang's `GameTest` system 🤔. That's all out of scope for this PR though.

**JUnit 5** is chosen as the testing toolkit to run the tests. [Upstream docs](https://junit.org/junit5/docs/current/user-guide/#writing-tests).

**Mockito** is chosen as a mocking framework to mock class dependencies. [Upstream docs](https://site.mockito.org/#how).

[**fabric-loader-junit**](https://github.com/FabricMC/fabric-loader/blob/0150462491e27810884fdaa45db96d6d35d9966a/junit/src/main/java/net/fabricmc/loader/impl/junit/FabricLoaderLauncherSessionListener.java) tells JUnit to use fabric's `Knot` classloader. This [enables things like](https://fabricmc.net/wiki/documentation:fabric_loader#class_loading_and_transformation) mixin & Environment stripping.

**See also** 

[Writing good tests](https://github.com/mockito/mockito/wiki/How-to-write-good-tests), in particular: [don't mock a type you don't own](https://github.com/mockito/mockito/wiki/How-to-write-good-tests#dont-mock-a-type-you-dont-own); a long term goal should be to reduce hard dependencies on mc classes & also write integration tests.

[AssertJ](https://assertj.github.io/doc/) (alternative assertions syntax we could use instead).

Fixes #151

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-name: Gradle build
+name: Gradle (build & test)
 
 on:
   push:
@@ -15,7 +15,7 @@ permissions:
   contents: read
 
 jobs:
-  build:
+  build_and_test:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
@@ -24,8 +24,12 @@ jobs:
       with:
         java-version: '17'
         distribution: microsoft
-    - name: Build with Gradle
+    - name: Build
       uses: gradle/gradle-build-action@v2
       with:
         arguments: build
-    # TODO: run gradle test
+    - name: Test
+      uses: gradle/gradle-build-action@v2
+      with:
+        # We could run `check` instead, but we may want to add checks we don't want to use in CI...?
+        arguments: test
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,31 @@
+name: Gradle build
+
+on:
+  push:
+    branches: &branches
+    - 'main'
+    - '1.16'
+    - '1.17'
+    - '1.18'
+    - '1.19'
+  pull_request:
+    branches: *branches
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up JDK
+      uses: actions/setup-java@v3
+      with:
+        java-version: '17'
+        distribution: microsoft
+    - name: Build with Gradle
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: build
+    # TODO: run gradle test
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Freecam
+![CI](https://github.com/MinecraftFreecam/Freecam/actions/workflows/main.yml/badge.svg?event=push)
 [![Crowdin](https://badges.crowdin.net/freecam/localized.svg)](https://crowdin.com/project/freecam)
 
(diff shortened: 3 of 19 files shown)


MattSturgeon:
Pinging @shartte as the original author of `BootstrapMinecraftExtension` & `@BootstrapMinecraft` to ask whether we can include them under the MIT license.

.github/workflows/main.yml
```diff
@@ -0,0 +1,35 @@
...
+on:
+  push:
+    branches: &branches
+    - 'main'
+    - '1.16'
+    - '1.17'
+    - '1.18'
+    - '1.19'
+  pull_request:
+    branches: *branches
```
MattSturgeon, line 12:
```
The workflow is not valid. .github/workflows/main.yml:
Anchors are not currently supported. Remove the anchor 'branches'
```

I guess GitHub doesn't like yaml anchors...
```suggestion
    branches:
    - 'main'
    - '1.16'
    - '1.17'
    - '1.18'
    - '1.19'
```

MattSturgeon:
Putting this comment from the Architectury discord here for my own reference...

> [My repository](https://github.com/Kotori316/FluidTank2) might help you.
> 
> It has both JUnit tests and `GameTest`, but no Mixin support as I don't use Mixin.
> [This file](https://github.com/Kotori316/special-octo-happiness/blob/b24d90d41cbd6ea388f9bb66b60a9d92cd80189d/test-utility/forge/src/main/java/com/kotori316/testutil/MCTestInitializer.java) also enables you to run JUnit test with Forge and NeoForge.
> I use `fabric-loader-junit` for fabric.
> It depends on Mockito to inject hacky codes.
> _\- [Kotori316](https://github.com/Kotori316)_

MattSturgeon:
Looks like https://github.com/neoforged/FancyModLoader/pull/54 will provide a Neoforge equivalent of `fabric-loader-junit`

.github/workflows/main.yml
MattSturgeon: For those who come after, GHA added basic yaml anchor support at some point.

PR #157, opened 2024-01-18
Fix disableNextTick preventing player from walking (regression)

### LINKED ISSUE #155: disableOnDamage freezes the player
When Freecam is deactivated because the player has suffered damage, you cannot move until you re-disable the Freecam.

Mods: Freecam 1.2.2 + Fabric API 0.93.1
Loader: Fabric 0.15.3 (Minecraft 1.20.4)
Modified Settings: allowInteract=true

MattSturgeon:
Regression introduced in 01143ecd54ea46480977154c525b77972de0546b because the `disableNextTick` check was moved before setting `mc.player.input`.

Fixes #155


hashalite (approved):
Looks good, thank you!

PR #162, opened 2024-02-01
Add a changelog & plugin

MattSturgeon:
## Changelog

This PR adds a changelog, generated from the GitHub releases' release notes.

The changelog was initially generated using [this bespoke tool](https://github.com/MattSturgeon/freecam-changelog-from-releases), however some other changes were made manually.

If there's anything that can be improved with the generated changelog it can be highlighted when reviewing this PR. For instance some release changes may be better placed in a different category. Other change lines may even benefit from being re-written to better fit the new format. I think I spotted one or two that are actually two changes on a single line; that may benefit from splitting into separate bullet points.

## Plugin

Additionally, a [changelog management plugin](https://github.com/JetBrains/gradle-changelog-plugin) by JetBrains is added. This plugin can be used to automate the process of bumping or "patching" the changelog when releasing a new version.

This is done by first ensuring `mod_version` is set to the new version being released, and then running `gradlew patchChangelog`. Anything in the "Unreleased" section of the changelog will then be marked as part of the current version and a new empty "unreleased" section will be generated.

### Plugin quirks

There's a few quirks of the plugin I should point out:

#### Link sorting

The bottom of the changelog defines the links used in each version header. Each time the plugin's patch task is run, it'll re-sort the links using it's [custom comparitor](https://github.com/JetBrains/gradle-changelog-plugin/blob/f055c31f493c47c6b2dfaf62c868bd000dad386b/src/main/kotlin/org/jetbrains/changelog/Changelog.kt#L147-L161).

For whatever reason, they've decided that [semver](https://semver.org/spec/v2.0.0.html) compliant versions should be sorted before other versions, so versions like `0.3` & `1.2.1.1` will always be at the end of the links because they do not have exactly three segments. 

#### Linked to tag

When patching, the plugin will generate a link to a diff between the `previous...new` releases. When doing this it assumes that each releases' tags exactly match the version name used in its changelog section header.

This is **incompatible** with the `1.2.2-mc1.20` style tags currently used.

Luckily, previous links will not be modified by the plugin, so this is only a consideration for future releases where `patchChangelog` will be used. Currently, I believe the best solution is to alias the "canonical" version of a release with a tag not containing the `-mc` part. For example, if we release `1.2.3-mc1.20`, `1.2.3.-mc1.19`, & `1.2.3-mc1.18`, then `1.2.3` should alias the "main" release, i.e. `1.2.3-mc1.20`.

While this would result in many _tags_ for each freecam version, we only need to create a single _release_ using one of those tags. We could add a link to the other tags in the release's notes & all artifacts could be added to one place.

This is great going forward, although it's not easy to backport to previous releases, because we would end up breaking GitHub's release date metadata and potentially end up publishing releases out of order too. That shouldn't be a major issue though because I've linked to each variant release under each version header in the changelog.

## Other PRs

It'd be good to have this merged before #140, because that PR can then make use of the plugin's "get changelog" feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -194,12 +194,13 @@ and Freecam's versioning is based on [Semantic Versioning](https://semver.org/sp
 - Made more strings translatable.
 - Updated Italian and Chinese translations (credits to @Loweredgames and @GodGun968).
 - Updated license to MIT in mod list screen.
+- Tripods are now stored separately per-dimension.
 
 ### Fixed
 
 - Fixed crash when enabling freecam whilst on fire or in lava with 'Disable on Damage' set to true.
 - Disable on damage now ignores damage received in creative mode.
-- Tripods are no-longer cleared when switching dimensions, and are now stored separately per-dimension.
+- Tripods are no-longer cleared when switching dimensions.
 - Fixed being able to hit tripods.
 
 ## [1.1.7] - 2022-12-10
@@ -533,9 +534,6 @@ and Freecam's versioning is based on [Semantic Versioning](https://semver.org/sp
 ### Changed
 
 - Holding jump and sneak at the same time will now make you hover in place.
-
-### Fixed
-
 - Player now remains sneaking if freecam is enabled while sneak is held.
 
 ## [0.4.8] - 2022-03-03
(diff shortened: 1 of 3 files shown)


hashalite (approved):
I've moved a few things in the changelog that seemed out of place. Everything else looks great, thank you!

PR #174, opened 2024-02-07
Generate lang files at build time

MattSturgeon:
- [ ] Refactor `:metadata` into a more general `:data` project
- [ ] Introduce a custom gradle task that can "build" lang files
- [ ] Move most `VariantTooltip` logic to the gradle task
- [ ] Move mod description to lang files
- [ ] Introduce gradle task/method to get English description from built lang files
- [ ] Reformat lang files as `.properties`


MattSturgeon:
This is feature complete and ready for review.

Notes:
- First time using Kotlin; seems nice so far. A very expressive language, and excellent null safety.
  - Kotlin was used to experiment with what it's like, with a view to potentially porting other parts of the build script in future.
- I would have liked to implement functional tests using Gradle's `TestKit`, however I was running into issues getting `buildSrc` into the runner's classpath...
  - I may still add some _unit_ tests though.
- I opted to port the translations from json to properties.
  - This is optional and can be reverted (with some work and changes to `LangTask`).
  - Crowdin _should_ handle this fine (untested).
  - Crowdin should be able to use comments added to the properties file as "context" for translations (untested).
  - You may have to manually tell your IDE to use `UTF-8` for properties files
    - If the chinese/japenese files look broken:
    - `Settings` -> `Editor` -> `File Encodings` -> `Default encoding for properties files`: `UTF-8`
    - This default setting _may_ be a good enough reason not to use properties files for translations...
- `buildSrc` is an entirely separate "build", with it's own gradle "projects".
  - Gradle automatically compiles `buildSrc` before running the main build, and makes the result available to the build script.
  - It is a simple way to have freecam-specific build logic, plugins, & tasks without having to clutter up the actual build script.
- `VariantTooltip` is now mostly a build step, the only thing handled at runtime is line `count`.
- Mod name and description are now handled by `:data`'s `langTask`.
  - If there is a `freecam.name.<variant>` translation, it is appended to the end of `freecam.name`.
  - Same for `description`.
  - These are also used for the name/description in the mod metadata file.
  - These are also assigned to the various `modmenu`-specific translation keys.
  - I'm not aware of any way to translate name/description on Forge (not yet investigated).
- I've verified that `:data` builds everything correctly.

data/build.gradle
```diff
@@ -0,0 +1,108 @@
...
+            // LangTask.getTranslation() can't be used at configuration time...
+            doLast {
+                def values = new TreeMap(inputs.properties)
+                values.mod_id = rootProject.name.toLowerCase()
+                values.name = langTask.get().getTranslation("freecam.name")
+                values.description = langTask.get().getTranslation("freecam.description")
+                values.json_authors = values.authors.split(',').collect { "\"$it\"" }.join(", ")
+
+                copy {
+                    // Include files under `src/platform`
```
MattSturgeon, line 75: Comment can be dropped or moved to `def inputDir`

data/src/lang/en-US/config.meta
```diff
@@ -0,0 +1 @@
+prefix=text.autoconfig.freecam.
```
MattSturgeon, line 1: Should the trailing `.` be assumed?

Prefix/suffix could be implemented as

```kotlin
listOfNonNull(prefix, key, suffix).joinToString(".")
```

buildSrc/src/main/kotlin/LangTask.kt
```diff
@@ -0,0 +1,145 @@
...
+    private fun readLangFile(file: File): Map<String, String> {
+        val metaFileName = file.name.replaceAfterLast('.', "meta")
+        val metaFile = inputDirectory.get().asFile.resolve(source.get()).resolve(metaFileName)
+        val meta = readPropertiesFile(metaFile)
+
+        return readPropertiesFile(file).mapKeys { (key, _) ->
+            var ret = key
+            meta["prefix"]?.let { ret = it + ret }
+            meta["suffix"]?.let { ret += it }
+            ret
```
MattSturgeon, line 121: See comment on `config.meta`; perhaps this should be implemented using:

```kotlin
listOfNonNull(prefix, key, suffix).joinToString(".")
```

buildSrc/src/main/kotlin/LangTask.kt
MattSturgeon, line 1: These classes are in the root package, which is probably bad practice. It does mean we don't have to import them in the build script though...

If we did put them in a package, then the other classes could be made package-private.

.github/workflows/crowdin-upload-translations.yaml
```diff
@@ -2,13 +2,12 @@ name: Crowdin (Upload Translations)
...
-#      - main
-#    paths:
-#      - '**/lang/*.json'
-#      - '!**/lang/en_us.json'
+  push: # Run when a non-en_us lang file is modified on the main branch
+    branches:
+      - main
+    paths:
+      - 'data/src/lang/**'
+      - '!data/src/lang/en-US/*'
```
MattSturgeon, line 10: It should be ok to auto-upload changes to translation files now. It's less likely we'll hit conflicts now that the download workflow is triggered by a webhook.

That said, we could still overwrite unmerged translations that are waiting in a PR... Maybe this should be reverted?

buildSrc/src/main/kotlin/LangTask.kt
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    fun readPropertiesFile(file: File): Map<String, String> {
+        val reader = try {
+            file.reader()
```
MattSturgeon, line 127: This reader should probably specify charset. Currently it returns a UTF-8 reader, however (for some reason) intellij defaults to opening  properties files using [ISO 8859-1]... Presumably for compatibility with legacy java applications?

[ISO 8859-1]: https://en.wikipedia.org/wiki/ISO/IEC_8859-1

MattSturgeon (commented):
Funny how you spot things you miss in the IDE when reading through it again on GitHub... Noted a few things so I remember to sort them.

Also noted a few things you may have an opinion on.

.github/workflows/crowdin-upload-translations.yaml
MattSturgeon: Reverted

buildSrc/src/main/kotlin/LangTask.kt
MattSturgeon: Moved to `net.xolt.freecam.gradle`

buildSrc/src/main/kotlin/LangTask.kt
MattSturgeon: Feature removed

buildSrc/src/main/kotlin/net/xolt/freecam/gradle/ModNameProcessor.kt
MattSturgeon, line 1: `ModNameProcessor` & `ModDescriptionProcessor` are implemented almost identically. The only difference is how ModMenu's "summary" is handled, which assumes the summary should be the description without any variant-specific suffix.  
We may prefer having separate `long_description` and `short_description` down the line?


There's probably a way to generalise them into a single implementation, but as I'm new to kotlin I'm not spending time designing that for the sake of two functions.

MattSturgeon:
Added some basic unit tests for the `LangProcessor` implementations & cleaned up `LangTask` a little.

buildSrc/src/main/kotlin/net/xolt/freecam/gradle/ModNameProcessor.kt
MattSturgeon: Maybe some kinda factory or builder would help here?

MattSturgeon:
Dropped the [forge mixin](https://github.com/MattSturgeon/Freecam/commit/754bae1fe13c9af872873c273513dbaa4cd94d20) for now, hopefully we'll get a proper upstream solution.

Issue #175, opened 2024-02-14
Freecam switches when closing gamemode switchet (F3+F4)


AlphaCraft9658:
When I use the gamemode switcher (F3+F4) and then release the F3 key, the freecam also switches on/off. I'm not sure whether F4 is the default key for the freecam, but at least with my configuration this happens. I don't think that this is intended behavior, and it should be possible to fix.

MattSturgeon:
I've reproduced this on Freecam 1.2.0 (old version) and also 1.2.3 (latest version).

This only happens when releasing `F3` **before** `F4`.

The _intended_ way to use the gamemode switcher is to **hold** `F3` like a _modifier key_ and then **tap** `F4` before releasing `F3`.

Notably, holding `F4` and tapping `F3` does not activate the gamemode switcher at all.

When used as intended, the gamemode switcher **does not** toggle Freecam.

It's also harder to reproduce this issue on the latest Freecam versions because #126 introduced a _half-second_ timeout for how long `F4` can be held before it is ignored.

If this issue is bothering you, there's a few workaround:

- Change Freecam's toggle key mapping
- Use gamemode switcher in the "intended" way (release `F4` first)
- Hold `F4` for at least half a second (10 ticks)

I'll look into a proper solution, but this might end up being marked as "won't fix".

AlphaCraft9658:
> I've reproduced this on Freecam 1.2.0 (old version) and also 1.2.3 (latest version).
> 
> This only happens when releasing `F3` **before** `F4`.
> 
> The _intended_ way to use the gamemode switcher is to **hold** `F3` like a _modifier key_ and then **tap** `F4` before releasing `F3`.
> 
> Notably, holding `F4` and tapping `F3` does not activate the gamemode switcher at all.
> 
> When used as intended, the gamemode switcher **does not** toggle Freecam.
> 
> It's also harder to reproduce this issue on the latest Freecam versions because #126 introduced a _half-second_ timeout for how long `F4` can be held before it is ignored.
> 
> If this issue is bothering you, there's a few workaround:
> 
> - Change Freecam's toggle key mapping
> - Use gamemode switcher in the "intended" way (release `F4` first)
> - Hold `F4` for at least half a second (10 ticks)
> 
> I'll look into a proper solution, but this might end up being marked as "won't fix".

For me it's kind of the opposite. When using the gamemode switcher normally, as you have described, it actuall still triggers the freecam for me.

MattSturgeon:
[...]

PR #182, opened 2024-02-28
Refactor combo key logic

MattSturgeon:
Hooking into vanilla's `KeyMapping.setDown()` allows use to greatly simplify the logic in `FreecamComboKeyMapping.tick()` without any change in behaviour.

This is just small improvements I made while investigating #175. An actual fix for that will be separate, if I can even come up with a good solution...

### DIFF
--- a/common/src/main/java/net/xolt/freecam/config/ModBindings.java
+++ b/common/src/main/java/net/xolt/freecam/config/ModBindings.java
@@ -7,11 +7,9 @@ import net.xolt.freecam.Freecam;
 import net.xolt.freecam.config.keys.FreecamKeyMapping;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.Objects;
-import java.util.Spliterator;
+import java.util.*;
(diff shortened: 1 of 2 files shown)


hashalite (approved):
Looks good, thank you!

common/src/main/java/net/xolt/freecam/config/ModBindings.java
```diff
@@ -80,4 +78,11 @@ public static void forEach(@NotNull Consumer<FreecamKeyMapping> action) {
                 .map(ModBindings::get)
                 .spliterator();
     }
+
+    /**
+     * Static implementation of {@link Collection#stream()}.
+     */
+    public static @NotNull Stream<FreecamKeyMapping> stream() {
+        return Arrays.stream(values()).map(ModBindings::get);
```
MattSturgeon, line 86: `forEach`, `itterator`, etc could make use of this method.

MattSturgeon: Done

hashalite (approved):

PR #188, opened 2024-03-10
Removed Operator Check

wallop560:
thats all i did. i want to use this on servers bro


MattSturgeon:
Freecam's Modrinth Edition has restrictions to comply with their policies. The builds available on GitHub and CurseForge are unrestricted.

This PR doesn't address your issue; instead you should just use a non-Modrinth version of Freecam.

PR #190, opened 2024-03-12
config: movement speed sliders

MattSturgeon:
Adds slider widgets for horizontal and vertical movement speed.

Introduces a custom `@BoundedScalable` annotation and `DoubleSliderEntry` GUI, based on cloth-config's [`@BoundedDiscrete`](https://shedaniel.gitbook.io/cloth-config/auto-config/annotations#configentry.boundeddiscrete) and [`IntegerSliderEntry`](https://github.com/shedaniel/cloth-config/blob/461ae8e56fa02f0422e2b8a3b1f38d743e5fba7e/common/src/main/java/me/shedaniel/clothconfig2/gui/entries/IntegerSliderEntry.java#L45) respectively.

Unlike my earlier implementation in #121, which ported the speed options to `int` so it could use cloth-config's `@BoundedDiscrete`, this implementation will not affect how speed values are stored or saved.

The PR is split into two parts; the first commit is a general refactor to allow adding more cloth-config extensions than just `VariantTooltip`. The second commit contains the meat of this PR.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and Freecam's versioning is based on [Semantic Versioning](https://semver.org/sp
 
 ### Changed
 
+- Movement speed options now use sliders instead of text fields.
+
 ### Removed
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and Freecam's versioning is based on [Semantic Versioning](https://semver.org/sp
 
 ### Changed
 
-- Movement speed options now use sliders instead of text fields.
+- Movement speed options now use sliders instead of text fields ([#190](https://github.com/MinecraftFreecam/Freecam/pull/190)).
 
 ### Removed
 
--- a/common/src/main/java/net/xolt/freecam/config/ModConfig.java
+++ b/common/src/main/java/net/xolt/freecam/config/ModConfig.java
@@ -8,6 +8,7 @@ import me.shedaniel.autoconfig.annotation.ConfigEntry.Gui.EnumHandler.EnumDispla
 import me.shedaniel.autoconfig.serializer.JanksonConfigSerializer;
 import me.shedaniel.clothconfig2.gui.entries.SelectionListEntry;
 import net.xolt.freecam.config.gui.AutoConfigExtensions;
+import net.xolt.freecam.config.gui.BoundedContinuous;
 import net.xolt.freecam.config.gui.VariantTooltip;
 import net.xolt.freecam.variant.api.BuildVariant;
 
@@ -32,9 +33,11 @@ public class ModConfig implements ConfigData {
         public FlightMode flightMode = FlightMode.DEFAULT;
 
         @ConfigEntry.Gui.Tooltip
+        @BoundedContinuous(max = 10)
         public double horizontalSpeed = 1.0;
 
         @ConfigEntry.Gui.Tooltip
+        @BoundedContinuous(max = 10)
         public double verticalSpeed = 1.0;
     }
 
--- a/common/src/main/java/net/xolt/freecam/config/ModConfig.java
+++ b/common/src/main/java/net/xolt/freecam/config/ModConfig.java
@@ -7,6 +7,8 @@ import me.shedaniel.autoconfig.annotation.ConfigEntry;
 import me.shedaniel.autoconfig.annotation.ConfigEntry.Gui.EnumHandler.EnumDisplayOption;
 import me.shedaniel.autoconfig.serializer.JanksonConfigSerializer;
 import me.shedaniel.clothconfig2.gui.entries.SelectionListEntry;
+import net.xolt.freecam.config.gui.AutoConfigExtensions;
+import net.xolt.freecam.config.gui.VariantTooltip;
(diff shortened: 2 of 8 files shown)


common/src/main/java/net/xolt/freecam/config/gui/DoubleSliderEntry.java
```diff
@@ -0,0 +1,150 @@
...
+        this.defaultValue = defaultValue;
+        this.maximum = maximum;
+        this.minimum = minimum;
+        this.precision = precision;
+        this.saveCallback = save;
+        this.sliderWidget = new Slider(0, 0, 152, 20, (this.value.get() - minimum) / (maximum - minimum));
+        this.sliderWidget.updateMessage();
+        this.resetButton = Button.builder(resetText, widget -> this.setValue(this.defaultValue.get()))
+                .width(MC.font.width(resetText) + 6)
+                .build();
```
MattSturgeon, line 53: I believe 1.19 is the earliest version with `Button.builder`. Older versions will need to use `new Button` instead.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and Freecam's versioning is based on [Semantic Versioning](https://semver.org/sp
 
 ### Changed
 
+- Movement speed options now use sliders instead of text fields.
```
MattSturgeon, line 14: Should this link to the PR?
```suggestion
- Movement speed options now use sliders instead of text fields ([#190](https://github.com/MinecraftFreecam/Freecam/pull/190)).
```

common/src/main/java/net/xolt/freecam/config/gui/BoundedContinuousImpl.java
```diff
@@ -0,0 +1,31 @@
...
+
+    static void apply(GuiRegistry registry) {
+
+        registry.registerAnnotationProvider(
+                (i18n, field, config, defaults, guiProvider) -> {
+                    Consumer<Double> save = newValue -> Utils.setUnsafely(field, config, newValue);
+                    Supplier<Double> defaultValue = () -> Utils.getUnsafely(field, defaults);
+                    double value = Utils.getUnsafely(field, config, defaultValue.get());
+                    BoundedContinuous bounds = field.getAnnotation(BoundedContinuous.class);
+                    return Collections.singletonList(new DoubleSliderEntry(Component.translatable(i18n), bounds.precision(), bounds.min(), bounds.max(), value, RESET_TEXT, defaultValue, save));
```
MattSturgeon, line 25: This would be more readable with a builder, but it probably isn't worth the effort it this is the only place we construct a `DoubleSliderEntry`.

common/src/main/java/net/xolt/freecam/config/ModConfig.java
```diff
@@ -30,9 +33,11 @@ public static class MovementConfig {
         public FlightMode flightMode = FlightMode.DEFAULT;
 
         @ConfigEntry.Gui.Tooltip
+        @BoundedContinuous(max = 10)
         public double horizontalSpeed = 1.0;
 
         @ConfigEntry.Gui.Tooltip
+        @BoundedContinuous(max = 10)
```
MattSturgeon, line 40: Not sure what the best "max" is; lower values provide less range but allow the user to be more precise, especially when setting low values.

common/src/main/java/net/xolt/freecam/config/gui/DoubleSliderEntry.java
```diff
@@ -0,0 +1,150 @@
...
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+import static net.xolt.freecam.Freecam.MC;
+
+class DoubleSliderEntry extends TooltipListEntry<Double> {
```
MattSturgeon, line 28:
```suggestion
/**
* @see IntSliderEntry
*/
class DoubleSliderEntry extends TooltipListEntry<Double> {
```

Should probably reference the "original" source for both attribution and future reference.

MattSturgeon:
Would it be worth the extra complexity to have _both_ a text-box and a slider? If so I might tackle that in a follow up PR once this is merged...

common/src/main/java/net/xolt/freecam/config/ModConfig.java
hashalite: 10 seems like a nice round number to me. 20 seems uncontrollable, even at extreme render distances. Maybe if an input box is added, it could bypass the maximum?

common/src/main/java/net/xolt/freecam/config/gui/BoundedContinuousImpl.java
hashalite: Seems fine as is

hashalite (approved):
Looks good, thank you!

common/src/main/java/net/xolt/freecam/config/ModConfig.java
MattSturgeon: I think 10 makes sense as is, but 5 might be better if we find a way to also have a text box.

PR #192, opened 2024-03-30
Version bumping task & CI

MattSturgeon:
This PR adds a version bumping gradle task, and a CI workflow that makes use of it.

[This workflow run](https://github.com/MattSturgeon/Freecam/actions/runs/8493999134) on my fork demonstrates the CI job working and [this commit](https://github.com/MattSturgeon/Freecam/commit/be81d3c80010f5226a0b9733320ffd56593b80a1) was pushed as a result.

### But, why?

Well, GitHub's branch protection prevents non-admins from pushing directly to a branch. This means I wouldn't be able to prepare a release commit (bumping version & changelog) without making a PR and getting it reviewed.

Luckily, branch protection doesn't apply to CI, so we can work around this by creating a workflow that is manually triggered (`workflow_dispatch`).

### Bonus

As a bonus, this gives all maintainers (even admins) a simple & unified way to do the mundane task of prepping a release's metadata.

We can also use `workflow_dispatch` to configure inputs, such as which version "level" is to be bumped.

image

### Future work

I'd like to add additional workflows that are triggered when tags get pushed.

One workflow for "primary" tags, which will simply un-draft the matching release:

```shell
gh release edit "${{ github.ref_name }}" --draft=false
```

And another that will trigger for "`+mc`" tags, that publish to github, cusrseforge & modrinth, using modpublisher.

Once https://github.com/firstdarkdev/modpublisher/pull/11 is done, the github release can be published as a draft, which means the tag doesn't need to exist.

Maybe it would also make sense for the bump-version workflow to create a tag, but for now I think tagging should be manual :thinking: 

### Also

While testing I noticed we didn't update the changelog config for the new `v` tag prefix. That is fixed by this PR.

### DIFF
--- a/.github/workflows/bump-version.yaml
+++ b/.github/workflows/bump-version.yaml
@@ -1,0 +1,86 @@
+name: Bump project version
+
+on:
+  workflow_dispatch:
+    inputs:
+      bump:
+        description: Version level to be bumped
+        required: true
+        default: minor
+        type: choice
+        options:
+          - major
+          - minor
+          - patch
+#      pre:
+#        description: 'pre-release version'
+#        required: false
+#        type: string
+
+# Only allow running one build job at a time to optimise cache hits
+concurrency:
+  group: builds
+
+# Grant GITHUB_TOKEN write access
+permissions:
+  contents: write
+
+env:
+  # sed expression to strip whitespace from property keys
+  # will not match comment lines
+  # prints result, so use with -n
+  replace_key_whitespace: 's/^\s*\([^=#][^=]*\)\s*=\s*/\1=/p'
+
+jobs:
+  bump_version:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
+          distribution: microsoft
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v3
+
+      - name: Get current metadata
+        id: old_props
+        run: sed -ne "$replace_key_whitespace" gradle.properties >> $GITHUB_OUTPUT
+
+      - name: Bump version
+        run: ./gradlew bumpVersion --${{ inputs.bump }}
+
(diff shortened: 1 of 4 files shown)


.github/workflows/bump-version.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+      - name: Bump version
+        run: ./gradlew bumpVersion --${{ inputs.bump }}
+
+      - name: Bump changelog
+        run: ./gradlew patchChangelog
+
+      - name: Get updated version info
+        id: new_version
+        run: ./gradlew printVersion --quiet >> $GITHUB_OUTPUT
```
MattSturgeon, line 58: We could to this using `grep` & `sed` (or `awk`) if we want to avoid unnecessary gradle runs in CI

```shell
$ grep mod_version gradle.properties | sed -e 's/^.*= *//'
1.2.3
```

Would mean we can drop the `./gradlew init` too

MattSturgeon: Done something similar, just using `sed`.

MattSturgeon:
Removed the `printVersion` task in favor of just reading the `gradle.properties` file.

This is done with the complex looking sed command:

```
sed -ne 's/^\s*\([^=#][^=]*\)\s*=\s*/\1=/p' gradle.properties
```

<details><summary>Explanation</summary>
<p>

`-n` means don't print, while `/p` means print result. This means only matching lines are printed, so we can filter out comments and blank lines without using `grep`.

`s/pattern/replacement/opts` is a standard substitution expression.

`^([^=#][^=]*)=` (whitespace `\s` removed for clarity) matches a property key, capturing the key name in group `.

`\1=` replaces the property key with one that doesn't have any whitespace, so that we can print them to `GITHUB_OUTPUTS`. `\1` is the captured key name.

</p>
</details> 

This has been tested on my fork: [action run](https://github.com/MattSturgeon/Freecam/actions/runs/8504437741), [resulting commit](https://github.com/MattSturgeon/Freecam/commit/0b99a960ce0f7f79a1fc6f631e0f3c32d0fea4e3).

hashalite (approved):

PR #194, opened 2024-04-02
Add a "build" CI workflow

MattSturgeon:
This PR adds a CI workflow that builds every commit pushed (including to PRs) and uploads the jars as artifacts.

No reason to wait for #153 to get some basics like this :smiley: 

This PR's own CI job [successfully ran here](https://github.com/MinecraftFreecam/Freecam/actions/runs/8530290526).

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,75 @@
+name: Build
+
+on:
+  push:
+
+# Only allow running one build job at a time to optimise cache hits
+concurrency:
+  group: builds
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Not sure if it's a good thing that an yet-unmerged workflow was able to run? Just thinking about security implications...

MattSturgeon:
It might make sense to do this as a build matrix:

- matrix can be computed from `enabled_platforms` in `gradle.properties`
- building one platform at a time would make build failures easier to diagnose
- uploading artifacts would be much simpler
- the build would _probably_ be slightly slower
- idk if it makes more sense to run the matrix in parallel (speed) or sequentially (better caching)

hashalite (approved):

PR #195, opened 2024-04-04
Add GitHub publishing and CI publishing

MattSturgeon:
## Summary

This PR consists of three parts:

#### 1. Configure modpublisher for GitHub Releases

Some recent Modpublisher PRs added support for configuring the github release tag separately from the "version". This allows use to use `1.2.3+mc1.20.4` as our curseforge/modrinth version, while publishing to a GitHub Release tagged as `v1.2.3`.

Support for marking the Release as "draft" was also added. This has the advantage of allowing a release to be created _before_ the tag exists, without GitHub automatically creating an (unsigned) tag.

Now, running `./gradlew publishMod` will also publish a draft release to GitHub.

Currently untested.

#### 2. Auto publish when pushing `+mc*` tags

This is pretty simple; automatically run `./gradlew publishMod` whenever a tag ending in `+mc*` is pushed to one of the release branches (`main` and `1.*`).

Currently untested.

#### 3. Publish GH releases when pushing "canonical" tags

This is still WIP, and may be postponed until a follow up PR.

The concept is simple; set `draft=false` on the github release when a "release tag" gets pushed.

However, we may also want to create the release if it doesn't exist yet? Alternatively, we might want a new modpublisher feature where the release is marked `draft` only if the tag doesn't exist yet. https://github.com/firstdarkdev/modpublisher/issues/16

Currently WIP & untested.

## Extra notes

See [Filter pattern cheat sheet](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet), because it turns out the `on` tag/branch/path filters aren't actually globs. It's some weird custom mash-up of regex & glob...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -160,6 +160,17 @@ subprojects {
             curseID = "557076"
             modrinthID = "XeEZ3fK2"
 
+            github {
+                // Extract our repo slug from the github URL
+                repo = rootProject.source_code_url.replaceFirst("^https?://github\\.com/", "")
+                // GitHub releases must be associated with a git tag
+                // We want this to be the canonical tag, not the annotated build tag
+                tag = "v${rootProject.mod_version}"
+                // Create a draft release, so that GitHub doesn't actually create the tag
+                // does not affect already published releases
+                draft = true
+            }
+
             // Format display name, e.g. "1.2.4 for MC 1.20.4 (fabric)"
             displayName = "${rootProject.mod_version} for MC ${rootProject.minecraft_version} (${project.name})"
             version = project.version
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -160,13 +160,24 @@ subprojects {
...
+                // GitHub releases must be associated with a git tag
+                // We want this to be the canonical tag, not the annotated build tag
+                tag = "v${rootProject.mod_version}"
+                // Create a draft release, so that GitHub doesn't actually create the tag
+                // does not affect already published releases
+                draft = true
+            }
+
             // Format display name "[Fabric] Freecam 1.2.2 for MC 1.20.4"
             displayName = "[${project.name.capitalize()}] ${rootProject.name.capitalize()} ${rootProject.mod_version} for MC ${rootProject.minecraft_version}"
```
MattSturgeon, line 175: I think the build display name (used on modrinth/curseforge) should be simplified.

It doesn't need to include "freecam", since that's obvious from the context (being on the freecam modpage).

Maybe `v1.2.3 for MC 1.20.4 (fabric)` would work? Or maybe `v1.2.3 (1.20.4 fabric)`?

This isn't really related to this PR, I was just thinking about it while in this part of the codebase.

hashalite: Yes I agree - it's redundant, not sure why I was doing that in the first place. `1.2.3 for MC 1.20.4 (fabric)` is my preference.

MattSturgeon: Addressed in https://github.com/MinecraftFreecam/Freecam/pull/218

hashalite (approved):

hashalite (approved):

PR #200, opened 2024-04-19
Update to 1.20.5

MattSturgeon:
- ~~Neoforge builds temporarily disabled.~~ See upstream PR https://github.com/neoforged/NeoForge/pull/794
- Neoforge PR maven added
- Bumped Gradle & JDK to 8.5 & 21
- Source/target compatibility still set to 17

Feel free to push to this branch.

Some changes here may apply to the older versions too.

We'll likely want to use a Java 21 JDK on **all branches** now, which might mean updating them to Gradle 8.6?


settings.gradle
```diff
@@ -10,7 +10,7 @@ pluginManagement {
 
 include("common")
 include("fabric")
-include("neoforge")
+//include("neoforge")
```
MattSturgeon, line 13: We should be able to use the neoforge's PR maven while waiting for it to be merged upstream: `https://prmaven.neoforged.net/NeoForge/pr794`

MattSturgeon:
Updating the `FreeCamera` entity and a couple mixins was enough to run the game and use Freecam in-game. I haven't tested very thoroughly though.

build.gradle
```diff
@@ -101,7 +108,7 @@ allprojects {
     // see http://yodaconditions.net/blog/fix-for-java-file-encoding-problems-with-gradle.html
     tasks.withType(JavaCompile) {
         options.encoding = "UTF-8"
-        options.release = 17
+        options.release = 21
```
MattSturgeon, line 111: should we also bump our source compatibility, or leave it at 17 to make backporting easier?

Come to think of if, maybe it should be set to 8 :thinking:

common/src/main/java/net/xolt/freecam/mixins/EntityRendererMixin.java
```diff
@@ -8,18 +13,12 @@
...
-import net.minecraft.network.chat.Component;
-import net.minecraft.world.entity.Entity;
-
 @Mixin(EntityRenderer.class)
 public class EntityRendererMixin {
 
     // Prevent rendering of nametag in inventory screen
     @Inject(method = "renderNameTag", at = @At("HEAD"), cancellable = true)
-    private void onRenderLabel(Entity entity, Component text, PoseStack matrices, MultiBufferSource vertexConsumers, int light, CallbackInfo ci) {
+    private void onRenderLabel(Entity entity, Component text, PoseStack poseStack, MultiBufferSource multiBufferSource, int light, float partialTick, CallbackInfo ci) {
```
MattSturgeon, line 21: Some of these param renames could be applied elsewhere and backported

common/src/main/java/net/xolt/freecam/util/FreeCamera.java
```diff
@@ -34,7 +36,10 @@ public class FreeCamera extends LocalPlayer {
                     FeatureFlagSet.of(),
                     null,
                     MC.getCurrentServer(),
-                    MC.screen)) {
+                    MC.screen,
+                    Collections.emptyMap(), //TODO cookies
+                    MC.gui.getChat().storeState(), //TODO state
+                    false)) {
```
MattSturgeon, line 42: Need to check if these are the correct values to use.

`MC.gui.getChat().storeState()` is used elsewhere in the MC codebase.

gradle.properties
```diff
@@ -26,24 +26,25 @@ release_type=release
 # https://maven.parchmentmc.org/org/parchmentmc/data
 # Forge dependencies use maven's version range spec:
 # https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html
-minecraft_version=1.20.4
-parchment_version=1.20.2-2023.12.10
+minecraft_version=1.20.5-rc1
+parchment_version=1.20.4-2024.02.25
```
MattSturgeon, line 30: parchment version bump can be backported, or merged early.

IMO waiting for a full release of 1.20.5 shouldn't be necessary to merge this to main, we can always bump again later.

common/src/main/java/net/xolt/freecam/util/FreeCamera.java
hashalite: Using an empty map for serverCookies should be fine since the networking code won't be used in the first place. Same deal with using `MC.gui.getChat().storeState()` for chatState

PR #204, opened 2024-04-29
Declare main classes API status

MattSturgeon:
Turns out mods integrating with or attempting to be compatible with Freecam may wish to compile against our classes and use our API.

e.g. FirstPersonModel [compiles against](https://github.com/tr7zw/FirstPersonModel/blob/1760a72e724d2cf7325c13485e487b699af02dcb/gradle-compose.yml#L20) Freecam v1.1.6 and uses the public API `Freecam.isEnabled()` that has (luckily) remained ABI compatible since before our initial commit.

It may be useful to explicitly state the API status of various public methods, even if for no reason other than to make us think twice before changing the public API.

I've only done this for the _main_ classes, `Freecam` and `FreeCamera` for now. Ideally, we'd declare the API status of all public classes.

One method I could only trace back to v0.2.2, which is the oldest release on GitHub and pre-dates the initial commit (AFAICT).

### DIFF
--- a/common/src/main/java/net/xolt/freecam/Freecam.java
+++ b/common/src/main/java/net/xolt/freecam/Freecam.java
@@ -17,6 +17,7 @@ import net.xolt.freecam.tripod.TripodSlot;
 import net.xolt.freecam.util.FreeCamera;
 import net.xolt.freecam.util.FreecamPosition;
 import net.xolt.freecam.variant.api.BuildVariant;
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
 
 public class Freecam {
@@ -33,6 +34,7 @@ public class Freecam {
     private static FreeCamera freeCamera;
     private static CameraType rememberedF5 = null;
 
+    @ApiStatus.Internal
     public static void preTick(Minecraft mc) {
         // Disable if the previous tick asked us to,
         // or Freecam is restricted on the current server
@@ -53,10 +55,12 @@ public class Freecam {
         }
     }
 
+    @ApiStatus.Internal
     public static void postTick(Minecraft mc) {
         ModBindings.forEach(Tickable::tick);
     }
 
+    @ApiStatus.Internal
     public static void onDisconnect() {
         if (isEnabled()) {
             toggle();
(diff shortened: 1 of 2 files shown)


common/src/main/java/net/xolt/freecam/Freecam.java
```diff
@@ -296,22 +310,29 @@ public static void moveToPlayer() {
...
         return freeCamera;
     }
 
+    @ApiStatus.AvailableSince("1.2.3")
     public static void disableNextTick() {
         disableNextTick = true;
     }
 
+    @ApiStatus.AvailableSince("0.2.2")
     public static boolean isEnabled() {
```
MattSturgeon, line 324: This may be older. 0.2.2 is the oldest jar I can find.

common/src/main/java/net/xolt/freecam/util/FreeCamera.java
```diff
@@ -18,12 +18,15 @@
...
+import org.jetbrains.annotations.ApiStatus;
 
 import java.util.Collections;
 import java.util.UUID;
 
 import static net.xolt.freecam.Freecam.MC;
 
+@ApiStatus.Internal
+@ApiStatus.AvailableSince("0.4.0")
 public class FreeCamera extends LocalPlayer {
```
MattSturgeon, line 30: This was renamed from `ClonePlayerEntity` in 0.4.0.

common/src/main/java/net/xolt/freecam/Freecam.java
```diff
@@ -296,22 +310,29 @@ public static void moveToPlayer() {
         );
     }
 
+    @ApiStatus.AvailableSince("0.4.0")
     public static FreeCamera getFreeCamera() {
         return freeCamera;
     }
 
+    @ApiStatus.AvailableSince("1.2.3")
     public static void disableNextTick() {
```
MattSturgeon, line 319: Maybe this should be experimental/internal?

common/src/main/java/net/xolt/freecam/Freecam.java
```diff
@@ -86,6 +92,7 @@ public static boolean resetTripodHandler() {
         return reset;
     }
 
+    @ApiStatus.AvailableSince("0.3.1")
     public static void toggle() {
```
MattSturgeon, line 96: This was `private` until 0.3.1.

hashalite (approved):
This is a good idea, thank you!

PR #207, opened 2024-05-03
Fix disconnect hook and singleplayer check

MattSturgeon:
Fixes a long-standing issue with the on-disconnect hook not being specific enough to disconnecting from a world. This was highlighted by an [incompatibility](https://github.com/MinecraftFreecam/Freecam/issues/206) with Lunar Client.

Also fixes a minor issue where hosted LAN worlds are not considered as "singleplayer"; instead we should check if an integrated server is running the world.

Here's some builds since the CI isn't triggering: [pr-207.zip](https://github.com/MinecraftFreecam/Freecam/files/15205647/pr-207.zip)


Itznicd:
It works! please add this fix into all future versions 🙏

hashalite (approved):

PR #217, opened 2024-06-03
CI fixes

MattSturgeon:
- **Enable `build` CI on PRs**
This should make the build workflow run on PRs too, as originally intended.\
It may end up running twice when pushing to PRs on non-fork branches though...

- **Use owner's PAT in bump-version CI**
To bypass branch protection, we can use a Personal Access Token generated on @hashalite's account. Should be added to the repo secrets as "`HASHALITE_PAT`".

- **Ensure bump-version commits as a bot**
This is the default anyway for the _committer_, but for _author_ [stefanzweifel/git-auto-commit-action](https://github.com/stefanzweifel/git-auto-commit-action) uses the last commit's author, which may or may not be the same as the CI's triggerer (`github.actor`).

### DIFF
--- a/.github/workflows/bump-version.yaml
+++ b/.github/workflows/bump-version.yaml
@@ -66,6 +66,11 @@ jobs:
           # We need to commit as a user to ensure CI is run on the commit
           # We need to push as an admin/owner to bypass branch-protection
           token: ${{ secrets.HASHALITE_PAT }}
+          # Ensure we commit as a bot, since we can't reliably get an author slug from `github.actor`
+          commit_user_name: github-actions[bot]
+          commit_user_email: 41898282+github-actions[bot]@users.noreply.github.com
+          commit_author: ${{ github.actor }} <41898282+github-actions[bot]@users.noreply.github.com>
+          # The commit message, specify the new version
           commit_message: Bump version to ${{ steps.new_props.outputs.mod_version }}
 
       - name: Summarize
--- a/.github/workflows/bump-version.yaml
+++ b/.github/workflows/bump-version.yaml
@@ -21,9 +21,9 @@ on:
 concurrency:
   group: builds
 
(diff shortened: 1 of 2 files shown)


.github/workflows/bump-version.yaml
```diff
@@ -63,6 +63,14 @@ jobs:
         id: commit
         uses: stefanzweifel/git-auto-commit-action@v5
         with:
+          # We need to commit as a user to ensure CI is run on the commit
+          # We need to push as an admin/owner to bypass branch-protection
+          token: ${{ secrets.HASHALITE_PAT }}
+          # Ensure we commit as a bot, since we can't reliably get an author slug from `github.actor`
+          commit_user_name: github-actions[bot]
+          commit_user_email: 41898282+github-actions[bot]@users.noreply.github.com
+          commit_author: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
```
MattSturgeon, line 72:
```suggestion
          commit_author: ${{ github.actor }} <41898282+github-actions[bot]@users.noreply.github.com>
```

This would attribute the author's name to the github _username_, which could be a decent compromise that doesn't involve looking up the `/users/<username>` endpoint?

It'd still use the bot email though, and it wouldn't use "display" names.

e.g. `--author "hashalite <41898282+github-actions[bot]@users.noreply.github.com>"`

hashalite:
> It may end up running twice when pushing to PRs on non-fork branches though...

Someone seems to have found a fix for this here:
https://github.com/zopefoundation/meta/pull/146

.github/workflows/bump-version.yaml
hashalite: Yes, this would be better

MattSturgeon:
Thanks for the feedback. This is ready for another review :smile:

hashalite (approved):

PR #252, opened 2024-11-02
Update to 1.21.3

JulianVennen:


gradle.properties
MattSturgeon, line 1: Can you also update `CHANGELOG.md` with

```diff

diff --git a/CHANGELOG.md b/CHANGELOG.md
index 13e117f..2f94486 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and Freecam's versioning is based on [Semantic Versioning](https://semver.org/sp
 
 ### Added
 
+- 1.21.3 Support ([#252](https://github.com/MinecraftFreecam/Freecam/pull/252)).
+
 ### Changed
 
 ### Removed

```

common/src/main/java/net/xolt/freecam/mixins/GuiMixin.java
```diff
@@ -34,7 +27,7 @@ private void onGetCameraPlayer(CallbackInfoReturnable<Player> cir) {
     // Don't render pumpkin overlay while Freecam is active
     @Inject(method = "renderTextureOverlay", at = @At("HEAD"), cancellable = true)
     private void onRenderTextureOverlay(GuiGraphics guiGraphics, ResourceLocation shaderLocation, float alpha, CallbackInfo ci) {
-        if (Freecam.isEnabled() && shaderLocation.equals(PUMPKIN_BLUR_LOCATION)) {
+        if (Freecam.isEnabled() && shaderLocation.equals(ResourceLocation.withDefaultNamespace("misc/pumpkinblur"))) {
```
MattSturgeon, line 30: Do you think we still need to check it is the pumpkin blur texture?

My thought process is that since the overlay is now data-driven it could be any texture (and be caused by any equipped item), but regardless of what the overlay is it should only apply to the player and not freecam's camera :thinking: 

```suggestion
        if (Freecam.isEnabled()) {
```

Thoughts?

common/src/main/java/net/xolt/freecam/mixins/GuiMixin.java
```diff
@@ -34,7 +27,7 @@ private void onGetCameraPlayer(CallbackInfoReturnable<Player> cir) {
     // Don't render pumpkin overlay while Freecam is active
```
MattSturgeon, line 27:
```suggestion
    // Don't render equipped-item overlays while Freecam is active
    // E.g. the pumpkin overlay
```
(if we agree this should apply to any equipped item overlay)

common/src/main/java/net/xolt/freecam/mixins/EntityRendererMixin.java
```diff
@@ -18,7 +18,7 @@ public class EntityRendererMixin {
 
     // Prevent rendering of nametag in inventory screen
     @Inject(method = "renderNameTag", at = @At("HEAD"), cancellable = true)
-    private void onRenderLabel(Entity entity, Component text, PoseStack poseStack, MultiBufferSource multiBufferSource, int light, float partialTick, CallbackInfo ci) {
+    private void onRenderLabel(EntityRenderState entityRenderState, Component component, PoseStack poseStack, MultiBufferSource multiBufferSource, int i, CallbackInfo ci) {
```
MattSturgeon, line 21:
```suggestion
    private void onRenderLabel(EntityRenderState renderState, Component component, PoseStack poseStack, MultiBufferSource multiBufferSource, int packedLightCoords, CallbackInfo ci) {
```

Looking at the parameter mappings for `Lnet/minecraft/client/gui/Font;drawInBatch` that `i` gets passed through to, we can use the name `packedLightCoords`.

common/src/main/java/net/xolt/freecam/mixins/LightTextureMixin.java
```diff
@@ -1,20 +1,21 @@
 package net.xolt.freecam.mixins;
 
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
```
MattSturgeon, line 4: Am I correct in believing that both fabric and neoforge bundle mixinextras these days? So as long as we don't use mixinextras in old version backports we are good to use them in new versions.

Do you know OTTOYH how far back mixinextras support goes for each platform?

MattSturgeon (commented):
Thanks for working on this! I had started on a port locally, but hadn't had chance yet to investigate some of the mixins that needed re-working.

I'll pull this locally to test it out. Other than some minor feedback (below) the diff LGTM).

gradle.properties
```diff
@@ -26,25 +26,25 @@ release_type=release
...
 fabric_supported_mc_versions=
 
-neoforge_version=21.0.139-beta
+neoforge_version=21.3.8-beta
 neoforge_pr=
-neoforge_mc_req=[1.21,)
+neoforge_mc_req=[1.21.2,)
 neoforge_loader_req=[1,)
-neoforge_req=[21.0.110-beta,)
+neoforge_req=[21.3.8-beta,)
```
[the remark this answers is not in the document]
JulianVennen: I haven't tested this specifically, but I see no reason why this shouldn't work.

common/src/main/java/net/xolt/freecam/mixins/GuiMixin.java
JulianVennen: Yeah, that probably makes sense

common/src/main/java/net/xolt/freecam/mixins/LightTextureMixin.java
JulianVennen: I think both of them ship it by default, but I don't know since when
When I added these, I thought it was already used somewhere else 😅

common/src/main/java/net/xolt/freecam/mixins/GuiMixin.java
MattSturgeon: Seems the equality check wasn't actually working anyway; I was still seeing the overlay when testing locally... I guess `.equals()` was doing reference-based equality.

Your latest version works correctly for me.

gradle.properties
```diff
@@ -26,25 +26,25 @@ release_type=release
 # https://maven.parchmentmc.org/org/parchmentmc/data
 # Forge dependencies use maven's version range spec:
 # https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html
-minecraft_version=1.21
+minecraft_version=1.21.3
 parchment_version=1.21-2024.07.07
 supported_mc_versions=
 
 fabric_loader_version=0.15.11
```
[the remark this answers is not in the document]
MattSturgeon: Looks like you've bumped `fabric_loader_req` by mistake instead of `fabric_loader_version`.

One relates to the "minimum requirement" that ends up in our mod metadata, while the other is the version we download to build against.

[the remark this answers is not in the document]
JulianVennen: Yeah, sorry I misunderstood what you meant. It's fixed now

common/src/main/java/net/xolt/freecam/mixins/LightTextureMixin.java
MattSturgeon: > I thought it was already used somewhere else

Maybe, though I don't recall using it. I do remember seeing an announcement a (few?) months back from neoforge saying they'd added support.

gradle.properties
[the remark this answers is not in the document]
MattSturgeon: TBH I don't know what the minimum req loadea to run fabric 1.21.2+ would be, but if we're unsure it's better to ere on the side of caution and not bump it...

If you know roughly what it should be, I'm happy for it to also be bumped as appropriate :grin:

MattSturgeon (approved):
Everything looks good. Testing locally I haven't found any issues yet.

MattSturgeon:
Thanks again for dealing with this! :rocket:

JulianVennen:
Thanks for the quick review ^^

PR #253, opened 2024-11-02
refactor: introduce a `DummyInput` wrapper

MattSturgeon:
`DummyInput` can be used to provide an empty `Input` instance that remembers some state from a previous `Input` instance.

This small refactor adds a `DummyInput` class that extends `Input` and takes over the responsibility for persisting the player's "is sneaking" state when the player is not actively being controlled.

- This shifts some responsibility away from the main `Freecam` class.
- This will make persisting additional state (jump, sprint?) easier.
- This will make porting to 1.21.3 easier, as the `Input` class has been refactored upstream.

Currently only sneak is persisted. In the future we may decide to persist additional state, such as sprint or jump.


MattSturgeon:
This is currently written for 1.21, I'll rebase it on main to target 1.21.3 and stash these changes in a temp branch as we'll need them for backporting to 1.21 and lower

PR #286, opened 2025-06-21
1.21.6 support

hashalite:


neoforge/src/main/java/net/xolt/freecam/forge/FreecamForge.java
```diff
@@ -33,7 +33,7 @@ public static void registerKeymappings(RegisterKeyMappingsEvent event) {
         ModBindings.forEach(event::register);
     }
 
-    @EventBusSubscriber(bus = EventBusSubscriber.Bus.GAME, value = Dist.CLIENT)
+    @EventBusSubscriber(value = Dist.CLIENT)
```
MattSturgeon, line 36: This looks suspicious. Do you know where this change is documented?

Without this, how does neoforge know that the outer class (`FreecamForge`) and this inner class (`FreecamForge.GlobalEventHandler`) have different event busses?

hashalite: Looking at [this](https://github.com/neoforged/NeoForge/pull/2346) it seemed like the bus no longer needs to be specified

MattSturgeon: Nice. So FML has abstracted away the concept of a user-specified event bus, and simply sends the right events based on the `@SubscribeEvent` annotation and method name?

In that case we should be able to delete `GlobalEventHandler` and move its content up into `FreecamForge`. Or have I misunderstood?

MattSturgeon: Yep: https://github.com/neoforged/FancyModLoader/pull/290

antisoma:
I've tested this and it appears to be working as expected in 1.21.6.

MattSturgeon:
> New changes require approval from someone other than MattSturgeon because they were the last pusher.

@hashalite you'll have to handle final approval/merge

hashalite:
I am actually unable to approve because I am the author. I should definitely loosen the branch protection rules lol

Issue #290, opened 2025-07-06
[neoforge] mod entrypoint not correctly marked as client-side only


Marius-jj:
### Description of the issue

When starting the server with several mods the server crashes, removing the Freecam mod, and the server is able to start. For server hosting I am using Apex hosting, if that matters.

Tried with the exact same setup in a single player world, and then the game works fine with the Freecam mod installed.

### Steps to reproduce

1. Start a server
2. Server crashes

### Log output

```shell

```

### Crash report

```shell
---- Minecraft Crash Report ----
// You should try our sister game, Minceraft!

Time: 2025-07-06 07:55:27
Description: Mod loading failures have occurred; consult the issue messages for more details

net.neoforged.neoforge.logging.CrashReportExtender$ModLoadingCrashException: Mod loading has failed


A detailed walkthrough of the error, its code path and all known details is as follows:
---------------------------------------------------------------------------------------

-- Head --
Thread: main
Stacktrace:
	at MC-BOOTSTRAP/fml_loader@4.0.38/net.neoforged.fml.common.asm.RuntimeDistCleaner.processClassWithFlags(RuntimeDistCleaner.java:60) ~[loader-4.0.38.jar%2360!/:4.0] {}
-- Mod loading issue for: freecam --
Details:
	Caused by 0: java.lang.ExceptionInInitializerError
		at TRANSFORMER/freecam@1.3.0+mc1.21/net.xolt.freecam.config.ModConfig.init(ModConfig.java:30) ~[freecam-neoforge-1.3.0+mc1.21.jar%23331!/:?] {re:classloading}
		at TRANSFORMER/freecam@1.3.0+mc1.21/net.xolt.freecam.forge.FreecamForge.<init>(FreecamForge.java:25) ~[freecam-neoforge-1.3.0+mc1.21.jar%23331!/:?] {re:classloading}
		at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62) ~[?:?] {}
		at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502) ~[?:?] {}
		at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486) ~[?:?] {}
[...]

PR #293, opened 2025-07-08
fix(neoforge): Mark the entrypoint as client-side

MattSturgeon:
While the EventBusSubscriber is already marked as client-side, the Mod entrypoint should also be marked as such.

See https://docs.neoforged.net/docs/concepts/sides
and https://docs.neoforged.net/docs/gettingstarted/modfiles/#javafml-and-mod

This should fix https://github.com/MinecraftFreecam/Freecam/issues/290 (not tested).


hashalite:
This does indeed fix #290; tested on a local server

hashalite (approved):

PR #292, opened 2025-07-08
Bump to 1.21.7

MattSturgeon:
- **build: bump parchment to 1.21.6-2025.06.29**
This commit can be kept on the 1.21.6 code.

- **build: update to 1.21.7**
Bump to 1.21.7. Changes seem minimal, no mappings conflicts, etc.

modmenu and cloth-config have not released updates yet, however both still work.

cloth-config shows warnings in neoforge because it is still using `@OnlyIn`, which is deprecated in favour of checking [`Level#isClientSide()`](https://docs.neoforged.net/docs/concepts/sides/#levelisclientside) and [`dist = Dist.CLIENT`](https://docs.neoforged.net/docs/concepts/sides/#fmlenvironmentdist). See https://github.com/neoforged/NeoForge/pull/2397. Incidentally, this is probably what's causing https://github.com/MinecraftFreecam/Freecam/issues/290, I'll investigate a fix separately to this PR.

PR #296, opened 2025-08-03
Bump to 1.21.8

403Denied:
Minimal changes once again, minor version update to add support for minecraft release 1.21.8

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -26,21 +26,21 @@ release_type=release
 # https://maven.parchmentmc.org/org/parchmentmc/data
 # Forge dependencies use maven's version range spec:
 # https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html
-minecraft_version=1.21.7
-parchment_version=1.21.6-2025.06.29
+minecraft_version=1.21.8
+parchment_version=1.21.8-2025.07.20
 supported_mc_versions=
 
 fabric_loader_version=0.16.14
-fabric_api_version=0.128.2+1.21.7
+fabric_api_version=0.130.0+1.21.8
 fabric_loader_req=>=0.14.0
-fabric_mc_req=>1.21.7-
+fabric_mc_req=>1.21.8-
 fabric_supported_mc_versions=
 
-neoforge_version=21.7.10-beta
+neoforge_version=21.8.25
 neoforge_pr=
-neoforge_mc_req=[1.21.7,)
+neoforge_mc_req=[1.21.8,)
 neoforge_loader_req=[1,)
-neoforge_req=[21.7.3-beta,)
+neoforge_req=[21.8.25,)
 neoforge_supported_mc_versions=
 
 # Other dependencies


gradle.properties
```diff
@@ -26,26 +26,26 @@ release_type=release
...
 neoforge_loader_req=[1,)
-neoforge_req=[21.7.3-beta,)
+neoforge_req=[21.8.25,)
 neoforge_supported_mc_versions=
 
 # Other dependencies
 # https://maven.terraformersmc.com/releases/com/terraformersmc/modmenu
 # https://mvnrepository.com/artifact/me.shedaniel.cloth/cloth-config?repo=architectury
 # FIXME
 modmenu_version=15.0.0-beta.3
```
MattSturgeon, line 50: This can be bumped now too, although it isn't strictly tied to 1.20.8 and would also be a valid bump on 1.20.6.

```suggestion
modmenu_version=15.0.0
```

Since we don't bundle modmenu in our jar, and this specific version is only used locally in the dev environment, it doesn't matter too much.

gradle.properties
```diff
@@ -26,26 +26,26 @@ release_type=release
...
 # https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html
-minecraft_version=1.21.7
-parchment_version=1.21.6-2025.06.29
+minecraft_version=1.21.8
+parchment_version=1.21.8-2025.07.20
 supported_mc_versions=
 
 fabric_loader_version=0.16.14
-fabric_api_version=0.128.2+1.21.7
+fabric_api_version=0.130.0+1.21.8
```
MattSturgeon, line 34: They now have:

```suggestion
fabric_api_version=0.133.4+1.21.8
```

gradle.properties
```diff
@@ -26,26 +26,26 @@ release_type=release
 # https://maven.parchmentmc.org/org/parchmentmc/data
 # Forge dependencies use maven's version range spec:
 # https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html
-minecraft_version=1.21.7
-parchment_version=1.21.6-2025.06.29
+minecraft_version=1.21.8
+parchment_version=1.21.8-2025.07.20
 supported_mc_versions=
 
 fabric_loader_version=0.16.14
```
MattSturgeon, line 33: We could bump the loader too:

```suggestion
fabric_loader_version=0.17.2
```

gradle.properties
MattSturgeon, line 1: Could you also update https://github.com/MinecraftFreecam/Freecam/blob/main/CHANGELOG.md ?

Adding an entry to the "unreleased" section:

```md
### Added

- 1.21.8 Support ([296](https://github.com/MinecraftFreecam/Freecam/pull/296))
```

Or maybe it shouln't say "added support", since we already kinda supported it.

Maybe it should instead be:

```md
### Changed

- Build against 1.21.8 ([296](https://github.com/MinecraftFreecam/Freecam/pull/296))
```

?

MattSturgeon (commented):
Thanks for the PR.

I'm on the fence about whether we should continue building against 1.21.6 for now, to have compatibility with .6-8... I guess it's fine. EDIT: we already bumped to .7 in #292 anyway...

But due to the delay in me reviewing, there are a few more things that can be bumped or bumped further:

PR #340, opened 2025-12-24
Add an Option to Outline to the Local Player Model

NicBelanger23:
Love this mod and figured I'd try to address a small gripe I have with it.

I often find that when flying around I can lose where my player is, and need to exit freecam to regain my bearings.

- Added a visual setting that enables a glowing outline on the local player while freecamming. I'd love to hear what you think!

This is my first experience with Fabric API and modding Minecraft in general, so please point out anything outside of best practice. Thanks!


MattSturgeon:
Sorry for the long delay responding to this PR.

For context, we've been busy doing major internal refactors including changing the i18n [translation file](https://github.com/MinecraftFreecam/Freecam/blob/main/i18n/src/main/en_US/config.toml) format, refactoring the config system into a separate [model](https://github.com/MinecraftFreecam/Freecam/blob/main/common/src/main/java/net/xolt/freecam/config/model/ModConfigDTO.java) and [interface](https://github.com/MinecraftFreecam/Freecam/blob/main/common/src/main/java/net/xolt/freecam/config/ModConfig.java), and porting from multiple MC-version branches to a single multi-version branch (using [stonecutter](https://stonecutter.kikugie.dev/)).

The feature itself sounds useful and the approach used seems simple enough to maintain.

Are you willing to rebase your PR on the latest `main` revision and update it to be compatible with [supported mc versions](https://github.com/MinecraftFreecam/Freecam/blob/main/stonecutter.settings.toml) (using stonecutter [directives](https://stonecutter.kikugie.dev/wiki/config/params))?

MattSturgeon:
You've ended up with 370 commits here. Did you maybe rebase main onto your pr instead of rebasing your pr onto main?

As you need to cleanup git history anyway, and refactor changes to fit the new codebase, it may be simpler to reset to main, manually reapply your diff, and create a fresh commit.

NicBelanger23:
Yeah I made a bit of a mess right before bed. (Don't git while half asleep). Gonna clean this up and force push over it

NicBelanger23:
Okay, got things cleaned up. 

What's changed: 

I've added the ability to bind a toggle to the glow outline. 
I've move over the translations into the new files
I've disabled the feature on pre-1.21.11 builds

common/src/main/java/net/xolt/freecam/Freecam.java
```diff
@@ -252,6 +253,7 @@ private static void onDisableFreecam() {
 
     private static void onEnable() {
         MC.smartCull = false;
+        outlineEnabled |= ModConfig.get().shouldAutoOutlinePlayer();
```
MattSturgeon: `|=` is a bitwize operation, essentially:
```java
outlineEnabled = outlineEnabled | ModConfig.get().shouldAutoOutlinePlayer();
```

`outlineEnabled` is a boolean, not a bit field, so this seems wrong.

[This stack overflow](https://stackoverflow.com/a/34274848) claims:

> when applied to a boolean variable "`|=`" will set it to `TRUE` the first time it encounters a truthy expression on the right side and will HOLD its `TRUE` value for all `|=` subsequent calls. Like a latch.

> Yep, the same holds in Java. But it worth noting that such OR operation `y|=expr` is not short-circuit (unlike `y = y || expr`), meaning that `expr` always evaluated.

i18n/src/main/en_US/config.toml
```diff
@@ -72,6 +72,9 @@ fullBright.tooltip = "Increases brightness while in freecam."
 showSubmersionFog.label = "Show Submersion Fog"
 showSubmersionFog.tooltip = "Whether you see a fog overlay underwater, in lava, or powdered snow."
 
+autoOutlinePlayer.label = "Auto Outline Player"
+autoOutlinePlayer.tooltip = "Automatically outline your player when entering freecam."
```
MattSturgeon: Personally, I don't think we need to think of this as explicitly "automatic".

I'd rename the internal variables `autoOutlinePlayer` -> `outlinePlayer` and rename the user facing text:
```suggestion
outlinePlayer.label = "Outline Player"
outlinePlayer.tooltip = "Apply an outline to your player while in Freecam."
```

or "while Freecam is enabled".

i18n/src/main/en_US/keybinds.toml
```diff
@@ -6,3 +6,4 @@ toggle = "Toggle Freecam"
 controlPlayer.toggle = "Control Player"
 tripod.reset = "Reset Tripod"
 config.open = "Config GUI"
+outline.toggle = "Toggle Player Outline"
```
MattSturgeon: Hm. I think this should be consistent with the corresponding option. Either rename this to:
```suggestion
outlinePlayer.toggle = "Toggle Outline Player"
```

Or rename the option itself to
```toml
outline.label = "Player Outline"
```
or
```toml
outlinePlayer.label = "Player Outline"
```

i18n/src/main/en_US/config.toml
MattSturgeon: Ah I see. You have two controls:
- The config option, controlls initial state when enabling Freecam
- The keybind, toggles current state while Freecam is active

I'd mis-interpreted the keybind as effectively toggling the config option itself.

So perhaps:
```suggestion
outlinePlayer.label = "Outline Player"
outlinePlayer."tooltip[0]" = "Apply an outline to your player when activating Freecam."
outlinePlayer."tooltip[1]" = "The outline can also be controlled using the \"Toggle Player Outline\" keybind."
```

?

(off-topic: reminds me, now we have TOML and `"""` multiline-strings, we could replace `[i]` tooltips with string-splitting)

cloth-config/src/main/java/net/xolt/freecam/clothconfig/ModConfigScreenFactory.java
```diff
@@ -300,13 +300,25 @@ private SubCategoryListEntry visualCategory(ConfigEntryBuilder entryBuilder) {
                 .setSaveConsumer(value -> config().visual.showSubmersion = value)
                 .build();
 
+        //? if >=1.21.11 {
```
MattSturgeon: The "glowing" effect was added in 16w02a (part of 1.9), so we could expand support to older versions if it's not too much work.

cloth-config/src/main/java/net/xolt/freecam/clothconfig/ModConfigScreenFactory.java
MattSturgeon, line 1: (picking a random file to comment on)

Could you add an entry to the "unreleased" part of the CHANGELOG, following the same format as other "added" entries?

If we're only adding this to 1.21.11+, the changelog should mention that too; e.g. "added an outline player feature, available on 1.21.11+ builds" or "not available on Minecraft versions before 1.21.11".

MattSturgeon: Aside: on that topic, I probably wont merge this until _after_ 1.4.0 stable is released (`main` is currently a 1.4.0 release candidate).

Maybe in the future we will separate release versioning from `main` to reduce friction here 🤔

common/src/main/java/net/xolt/freecam/mixins/LevelRendererMixin.java
```diff
@@ -0,0 +1,46 @@
...
+import net.minecraft.client.DeltaTracker;
+import net.minecraft.client.renderer.culling.Frustum;
+import net.minecraft.client.renderer.entity.EntityRenderDispatcher;
+import net.minecraft.client.renderer.entity.state.EntityRenderState;
+
+//? if >=26.1
+import net.minecraft.client.renderer.state.level.LevelRenderState;
+
+//? if =1.21.11
+//import net.minecraft.client.renderer.state.LevelRenderState;
```
MattSturgeon: This looks a bit messy, especially nested within the outer `//? if >=1.21.11 {`

It could be simplified to `if/else` and the whitespace could be rearanged; currently it looks like the second `if` applies to the whole block through to L21.

Further, instead of `if/else` or `if/if`, we could actually use stonecutter's new "replace" feature:


```suggestion
//~ if >=26.1: 'state' -> 'state.level'
import net.minecraft.client.renderer.state.level.LevelRenderState;

```

__However__, turns out this is already handled by `stonecutter.gradle.kts`'s replacements:
https://github.com/MinecraftFreecam/Freecam/blob/f2caff599b3e8c060b21aa769267b8f07c3d6ab4/stonecutter.gradle.kts#L72


```suggestion
import net.minecraft.client.renderer.state.level.LevelRenderState;
```

common/src/main/resources/freecam-common.mixins.json5
```diff
@@ -15,6 +15,8 @@
     "GameRendererMixin",
     "GuiMixin",
     "IrisHandRendererMixin",
+    //? if >=1.21.11
+    "LevelRendererMixin",
```
MattSturgeon: IIUC, an empty Mixin class is effectively a no-op, so we _probably_ don't need to make it conditional here. WDYT?

MattSturgeon: I'd also like this list to stay alphabetical, so moving `LevelRendererMixin` down to between `ItemInHandRendererMixin` and `LivingEntityMixin`.

common/src/main/java/net/xolt/freecam/Freecam.java
MattSturgeon: Having tested it out in the IDE, still think this latch may be a bit confusing.

As is, the "toggled" state will be remembered until the client is completely closed; even exiting a world will remember state.

In my testing it is more intuitive for the state to _reset_ to the "enable outline on enable" value every time Freecam is enabled:

```suggestion
        outlineEnabled = ModConfig.get().shouldAutoOutlinePlayer();
```

i18n/src/main/en_US/config.toml
MattSturgeon: Maybe `shouldAutoOutlinePlayer` should be `shouldInitiallyOutlinePlayer` and the actual field could be `initiallyOutline` or `initiallyOutlinePlayer`?

We could also consider rephrasing "outline" as "highlight"... 🤔

MattSturgeon (commented):
Initial read through the diff. Some minor feedback, but seems good overall. Thanks for updating for our current codebase!

NicBelanger23:
I decided not to enable this feature in pre 1.21.11 versions. I don't think it's impossible to implement, however I think it will be quite a bit more complex. The rendering systems for glowing and shadows seems to have changed quite a bit over the previous versions and I don't have the know-how to make it work.

common/src/main/java/net/xolt/freecam/mixins/LevelRendererMixin.java
MattSturgeon: > **However**, turns out this is already handled by `stonecutter.gradle.kts`'s replacements:
> https://github.com/MinecraftFreecam/Freecam/blob/f2caff599b3e8c060b21aa769267b8f07c3d6ab4/stonecutter.gradle.kts#L72

Did you see this bit? Sorry I buried it a bit among other text.

MattSturgeon:
> CI failure

Looks like poor timing, one of the maven repos we need is down: https://status.trans.fish/status/kiku

cloth-config/src/main/java/net/xolt/freecam/clothconfig/ModConfigScreenFactory.java
MattSturgeon: > I decided not to enable this feature in pre 1.21.11 versions. I don't think it's impossible to implement, however I think it will be quite a bit more complex. The rendering systems for glowing and shadows seems to have changed quite a bit over the previous versions and I don't have the know-how to make it work.

That's fair. Do you mind adding a TODO comment with some context?

```java
// TODO: backport player outline to older versions
// See https://github.com/MinecraftFreecam/Freecam/pull/340
```

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and Freecam's versioning is based on [Semantic Versioning](https://semver.org/sp
 ## [Unreleased]
 
 ### Added
+- 'Player Glow' keybind that toggles a glowing outline around your player while in freecam mode
+  - 'Player Glow' setting that controls whether the outline is enabled when entering freecam
+  - English and Ukrainian translations for new features
+  - Only available in 1.21.11 and later due to mixin problems
```
MattSturgeon:
```suggestion
- A glowing player outline feature, to highlight your player while in Freecam ([340](https://github.com/MinecraftFreecam/Freecam/pull/340))
  - 'Outline Player' setting that to control whether the outline is initially enabled when entering Freecam (off by default)
  - 'Toggle Outline Player' keybind to enable or disable the outline while in Freecam
  - English and Ukrainian translations
  - For now, it is only available in Minecraft 1.21.11 and newer
```

MattSturgeon (approved):
I've rebased, ported to 26.2, and made a few minor tweaks.[^1]

I'll merge shortly and this should make it into `1.4.1-beta.2`, which I won't rush to release as I only did `beta.1` today.

Thanks for your work on this!

[^1]: This was a force-push, so you won't be able to `pull` normally. I normally do a `git fetch --all --prune` then a `git reset --hard @{u}` to reset to the remote's revision.

PR #388, opened 2026-03-31
build(meta): derive ReleaseType from semantic version

MattSturgeon:
Automatically derive the ReleaseType from the (semantic) version, as follows:
- `1.0.0-alpha.1` → `ALPHA`
- `1.0.0-beta.1` → `BETA`
- `1.0.0-rc.1` → `RELEASE_CANDIDATE`
- `1.0.0` → `RELEASE`

Also, adds a `:patchChangelog` reminder comment in `metadata.toml` and adds `ReleaseType.RELEASE_CANDIDATE` (`rc`) support.

Also collapsed the only changelog entry using `-beta1` format, and updated the changelog regex to expect `-beta` or `-beta.1` instead.

### DIFF
--- a/build-logic/api/src/main/kotlin/net/xolt/freecam/model/ReleaseType.kt
+++ b/build-logic/api/src/main/kotlin/net/xolt/freecam/model/ReleaseType.kt
@@ -1,7 +1,12 @@
 package net.xolt.freecam.model
 
+import io.github.z4kn4fein.semver.Version
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.decodeFromJsonElement
+import kotlinx.serialization.json.encodeToJsonElement
 
 @Serializable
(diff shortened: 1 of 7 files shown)


build-logic/api/src/main/kotlin/net/xolt/freecam/model/ReleaseType.kt
```diff
@@ -1,11 +1,41 @@
...
+}
+
+/** Extract a [ReleaseType] from a semantic version string. */
+fun ReleaseType.Companion.fromVersion(version: String, strict: Boolean = false): ReleaseType =
+    Version.parse(version, strict).toReleaseType()
+
+/** Extract a [ReleaseType] from a [Version]. */
+internal fun Version.toReleaseType(): ReleaseType =
+    preRelease
+        ?.substringBefore('.')
```
MattSturgeon: I'm assuming `alpha.123` rather than `alpha-123` here, as that seems to be the convention.

If we'd rather align with minecraft's new version style (`snapshot-1`, `pre-1`, `rc-1`), we could support that too/instead.

PR #390, opened 2026-04-01
Generate i18n files from TOML sources; support mod description

### LINKED ISSUE #172: i18n: Support translating mod description
#### Request

This topic was [brought up](https://crowdin.com/project/freecam/discussions/2) in Crowdin discussions by [Әлмәт Ак Арыслан
Әлмәт Ак Арыслан](https://crowdin.com/profile/19082004amir).

> Hello! I think it would be nice to have a Possibility to translate Description and Summary for ModMenu... It really helps when navigating through Content of premade Modpacks and also it just improves Mod's Accessibility.

#### Implementation

While [`fabric.mod.json`](https://fabricmc.net/wiki/documentation:fabric_mod_json_spec) doesn't natively support translated descriptions, modmenu [documents](https://github.com/TerraformersMC/ModMenu/wiki/API#translatable-metadata) that additional keys can be added to our language files to handle this. 

That may be more complex that it sounds, because we need a different description for the standard & Modrinth builds.

I suspect, we'll need to have the `:metadata` project generate lang files in addition to what it currently does.

This will mean moving the lang files out of `common/src/resources` and potentially adding additional lang files that we can merge when building.

We'll also need to move the description out of `gradle.properties` and into one of these new metadata language files.

This presents an opertunity to handle converting from crowdin's supported language keys without needing to use `languages_mappings`.

#### Off topic

I noticed while browsing the [modmenu docs](https://github.com/TerraformersMC/ModMenu/wiki/API#custom-mod-menu-metadata) that they support setting a number of extra things in a `"custom"` object in `fabric.mod.json`, some of these may be worth looking into.

MattSturgeon:
### Motivation

This PR introduces a structured i18n pipeline with two primary goals:

1. **Decouple translation keys from implementation details**
   - Removes legacy autoconfig naming
   - Allows consistent, structured key hierarchies
   - Avoids ambiguous "node + leaf" key collisions

2. **Enable deterministic generation and transformation**
   - Support mapping a single source key to multiple output keys
     - e.g. `mod.description` → `fml`, `modmenu`
   - Allow logical grouping via TOML (per-file, per-table)
   - Improve maintainability in-repo and in Crowdin

### Risk

This PR performs a full rewrite of:
- Translation key names
- File structure
- Generation pipeline

As a result:
- Existing Crowdin mappings will be invalidated
- All translations must be re-uploaded

Mitigation:
- Changes are split into logical commits for review
- `[TEMP]` commits document migration scripts and will be dropped before merge
- Output is deterministic and covered by tests

### Why this is a single PR

Either of the following changes would each require a full Crowdin re-upload:
- Renaming keys
- Changing file structure

Splitting this PR would therefore:
- Increase deployment overhead
- Risk inconsistent intermediate states

For that reason, this is intentionally atomic.

Review strategy:
- Review commit-by-commit
- Skip `[TEMP]` commits; they will be dropped before merge

### Design

- **Gradle plugin + task**
  - Task performs conversion and transformation
  - Plugin wires inputs/outputs and exposes configuration
- **Dedicated `:i18n` module**
  - Minimal dependencies
  - Built once and reused across all variants
  - Cacheable output
- **TOML-based source format**
  - Multiple files per locale
  - Structured via tables
  - Flattened into JSON at build time

- `meta.description` now sourced from `:i18n`
  - Lazily parses `:i18n/src/main/en_US/metadata.toml`
  - Separate from the i18n plugin

### Migration

Migration was performed in two stages:

1. **Key renaming**
   - Driven by a `mapping.json` (`old.key` → `new.key`)
   - Applied tree-wide by `remap.py`
2. **TOML restructuring**
   - Keys grouped by topic and table
   - Ordered to match approximate in-game UI layout
   - Applied by `migrate.py` and `migrate.fish`

Temporary scripts used for migration are included in `[TEMP]` commits for reference and will be removed before merge.

### Updating Crowdin

Post-merge:
1. CI automatically uploads `i18n/src/main/en_US`
2. Manually trigger the [upload translations] workflow

Follow-up:
- Verify old keys are removed from Crowdin
- Confirm no orphaned translations remain
- Fall-back to the `crowdin` [CLI][crowdin-cli] if needed

[upload translations]: https://github.com/MinecraftFreecam/Freecam/actions/workflows/crowdin-upload-translations.yaml
[crowdin-cli]: https://crowdin.github.io/crowdin-cli/

## Checklist
- [x] Implement i18n gradle plugin
- [x] Implement i18n format conversion task
- [x] Rename existing translation keys
- [x] Migrate existing dicts to new system
- [ ] Add CI to lint or validate translation files?
- [x] Move `meta.description` into i18n
      Closes #172
- [ ] Decide whether to add a CHANGELOG entry...?
      Not really user-facing, unless the user actively contributes translations
- [ ] Investigate how much of this can be offloaded to Fletching Table's [`LanguageFileTransformer`](https://codeberg.org/stonecutter/fletching-table/src/commit/45f694597f646bd2928d7b28c0a6228c122fcfd6/src/main/kotlin/dev/kikugie/fletching_table/transformer/LanguageFileTransformer.kt)
- [x] **Drop `[TEMP]` commits!**
- [x] Upload to Crowdin _(post-merge)_


build-logic/i18n-plugin/src/main/kotlin/net/xolt/freecam/gradle/I18nExtension.kt
```diff
@@ -0,0 +1,110 @@
...
+        generatedDir.file(lang.asMinecraftLocale + ".json").map { file ->
+            file.asFile.inputStream().use {
+                Json.decodeFromStream<Map<String, String>>(it)
+            }
+        }
+
+    /**
+     * Retrieve a translated value from the [default][defaultLanguage] language, by its [key] path.
+     */
+    operator fun get(key: String): Provider<String> = value(key)
```
MattSturgeon: These `i18n.value("key")` and `i18n["key"]` getters were originally implemented for an earlier iteration, when I hoped to replace `meta.description` with a task-backed `Provider`.

Turned out that worked fine with the `:fabric:generateModJson` task, which can accept a provider, but the `:forge:processResources` property expansion _cannot_ use a task-backed provider.

As `description` is currently the only translatable property we need to access outside of lang files, and we only need the english translation, I instead converted `meta.description` itself to read from `en_US/metadata.toml`.

Therefore, these getters are currently unused and could be dropped from the impl, if we don't think they will be useful or worth maintaining.

hashalite (dismissed):

MattSturgeon:
Thanks for the review!

I've rebased:
- re-applied migration to new translations on `main`
- dropped temporary commits
- added a changelog entry for translatable mod description

Not sure whether the (mostly internal) "renamed translation keys" is worth a changelog entry. Going with "no" for now.

PR #443, opened 2026-04-13
Replace `:publish` with `gh` CLI

MattSturgeon:
#### ci(publish): replace 'publisher' with 'gh'

GitHub's CLI is more than capable of atomically tagging and publishing a GitHub Release. We don't need a custom publishing tool for that.

We can read our release metadata using `jq` and publish the GH Release using `gh`.

#### remove(publish): drop

In its current form, it is redundant and can be replaced by `gh`.

In the future, we may reuse its codebase for CurseForge & Modrinth.
But when that happens it'll be maintained as a separate repo.

## Note

I've already extracted `:publish` to [another repo](https://github.com/MinecraftFreecam/Publisher), with git history. I will try to work on refactoring the freecam publisher to focus on CurseForge & Modrinth, and possibly also be a home for reusable CI Actions and Workflows related to publishing.

_Finally back to mostly Java:_

image

Closes #365 (needs porting to the new repo)
Partially reverts #354


MattSturgeon:
Tested triggering a release on my fork:

https://github.com/MattSturgeon/Freecam/releases/tag/v1.4.0-alpha.3

image

PR #527, opened 2026-06-26
ci(prepare): update untrusted checkout for v7

MattSturgeon:
actions/checkout v7 now requires explicitly defining `allow-unsafe-pr-checkout`
when checking out untrusted PR files in a privileged context.


MattSturgeon:
Bypassing CI as `pull_request_target` runs the base branch's workflow; changes here won't show up until merged.

PR #528, opened 2026-06-26
Port to 26.3

MattSturgeon:
- https://www.minecraft.net/en-us/article/minecraft-java-edition-26-3
- https://minecraft.wiki/w/Java_Edition_26.3

This cycle, Minecraft has ported from GLFW to SDL, and refactored hand attack/swing rendering to use the newer "extract render state" architecture. Most other changes have not affected us.

The PR includes some refactoring that allowed disabling the cloth-config build, targeting a NeoForge PR, and overriding a "published requirement" to `NONE`. These are not needed in the final state of the PR, but are included as separate cleanup that'll be useful in future updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,18 @@ and Freecam's versioning is based on [Semantic Versioning](https://semver.org/sp
 
 ## [Unreleased]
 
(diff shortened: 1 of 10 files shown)


MattSturgeon:
snapshot-3 will need a new fabric-api, as some of its mixins don't apply

PR #537, opened 2026-07-07
feat(fabric): drop fabric-api dependency

MattSturgeon:
Bundle the required fabric-api modules jar-in-jar and drop explicit dependency on `fabric-api: *`.

I had thought this wasn't possible, because `cloth-config` specifies a dependency on `fabric-api` itself (rather than a specific module). However, in testing it seems to work fine.
Maybe fabric-loader doesn't actually check the requirements of bundled jar-in-jar mods? I tested this with a few fabric-loader versions, just to be sure, and found no issues (back to fabricloader 1.14.0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ and Freecam's versioning is based on [Semantic Versioning](https://semver.org/sp
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


MattSturgeon:
This makes the jar ~200kB bigger, it goes from 1.1MB to 1.3MB. I believe the majority of that is still `cloth-config`.

I'd say on balance, the simpler installation requirements make 200kB worth it.

PR #547, opened 2026-07-22
Freecam disable packet

### LINKED ISSUE #546: Add ability for servers to disable freecam
### Description of the feature

A plugin message that can be sent to users of the mod to disable freecam.

### Why is this useful?

We as server owners don't want to ban people just for having freecam installed, its quite a popular mod, but we also don't want people using it to X-ray and gain an unfair advantage.

All the major mini-map (journey, xaeros, voxelmap etc.) mods have this already, server can send plugin message containing a payload to disable caves view or entities view.

If you are worried about this making the mod detectable, you don't have to make it send any plugin messages, just silently change the settings when it receives a specific message from the server.

### Freecam version

1.4.1

personalized-advertising:
Fixes #546
Not the cleanest possible integration, but should work.
Needs proper testing.

### DIFF
--- a/common/src/main/java/net/xolt/freecam/Freecam.java
+++ b/common/src/main/java/net/xolt/freecam/Freecam.java
@@ -31,11 +31,13 @@ public class Freecam {
 
     public static final Minecraft MC = Minecraft.getInstance();
     public static final String MOD_ID = "freecam";
+    public static final String SERVER_DISABLE_CHANNEL = "freecam:disable";
 
     private static boolean freecamEnabled = false;
     private static boolean tripodEnabled = false;
     private static boolean playerControlEnabled = false;
     private static boolean disableNextTick = false;
+    private static boolean serverDisabled = false;
     private static boolean outlineEnabled = false;
     private static final TripodRegistry tripods = new TripodRegistry();
     private static TripodSlot activeTripod = TripodSlot.NONE;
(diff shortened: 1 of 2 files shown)


MattSturgeon:
Thanks. This has been on my radar for a while now, I'll take a look later.

Long term, I'd like a Freecam server-side mod that reads a config file and informs the client of more granular "policies". That way, servers can configure things like "don't allow ignoring collision" or "don't allow interaction mode `CAMERA`" while still allowing other features.

Short term, we could recognise custom packet payloads for the features we think are likely to be controversial. But I'm also hesitant to introduce something we'll later deprecate once we have a new approach... 🤔

personalized-advertising:
> Thanks. This has been on my radar for a while now, I'll take a look later.
> 
> Long term, I'd like a Freecam server-side mod that reads a config file and informs the client of more granular "policies". That way, servers can configure things like "don't allow ignoring collision" or "don't allow interaction mode `CAMERA`" while still allowing other features.
> 
> Short term, we could recognise custom packet payloads for the features we think are likely to be controversial. But I'm also hesitant to introduce something we'll later deprecate once we have a new approach... 🤔

Does sound more complex than the server just saying "I don't want freecam here"

personalized-advertising:
> Thanks. This has been on my radar for a while now, I'll take a look later.
> 
> Long term, I'd like a Freecam server-side mod that reads a config file and informs the client of more granular "policies". That way, servers can configure things like "don't allow ignoring collision" or "don't allow interaction mode `CAMERA`" while still allowing other features.
> 
> Short term, we could recognise custom packet payloads for the features we think are likely to be controversial. But I'm also hesitant to introduce something we'll later deprecate once we have a new approach... 🤔

I will work on the client-side aspect of this and the network protocol, so it can be added now and then later used for the future server-side mod

personalized-advertising:
Closed in favor of #548

PR #564, opened 2026-07-28
build(loaders): fix issues with declarative source/resource dependencies

MattSturgeon:
This PR resolves Configure on Demand and project lifecycle issues across the common->loader dependency graph, by replacing eager project task evaluation with idiomatic output-configurations. This reveals a latent issue with Fabric Loom's repository mutations, which we also workaround.

We previously avoided using output-configurations because we ran into recursive project resolution issues when `group` was defined at the project top-level. That's since been resolved, so we should do things correctly.

This resolves issues that often get masked by Gradle cache, that I noticed when adding the 26.3 version and noticing that its `processResources` task was not run correctly.

### Changes
1. **Loom Repository Workaround (`freecam.common`):**
   - Pre-emptively registers the `MavenCentralLWJGL` repository to prevent Fabric Loom from attempting runtime repository mutations that trigger Gradle immutability lock errors.

2. **Configuration-Based Source Sharing (`freecam.loaders` & `:common`):**
   - Replaced fragile, eager `commonNode.project.tasks` lookups in `sourceSets.main` with dedicated consumable and resolvable configurations (`generatedSourcesElements` / `processedResourcesElements`).
   - Ensures Gradle lazily wires task execution dependencies (`builtBy`) and correctly preserves Configure on Demand performance for inactive version branches.

3. **Use task-generated Loom AW file where possible:**
   - I noticed when testing the above changes that the eval-time written AW file was causing issues for some loom tasks.
   - Where possible, we explicitly wire-in the task-generated file instead.

### DIFF
--- a/build-logic/conventions/src/main/kotlin/freecam.loaders.gradle.kts
+++ b/build-logic/conventions/src/main/kotlin/freecam.loaders.gradle.kts
@@ -4,15 +4,32 @@ plugins {
     id("freecam.common")
 }
 
+val extraJavaSources = configurations.create("extraJavaSources") {
+    description = "Additional Java sources added to the main source set."
+    isCanBeConsumed = false
+    isCanBeResolved = true
+}
+
+val extraResources = configurations.create("extraResources") {
+    description = "Additional resources added to the main source set."
+    isCanBeConsumed = false
+    isCanBeResolved = true
+}
+
 sourceSets {
     main {
-        // Manually depend on common's pre-processed sources.
-        // NOTE: loaders have no build dependency on common, so API/implementation
-        //  classes and transitive dependencies must be manually propagated.
-        val commonTasks = commonNode.project.tasks
-        java.srcDirs(commonTasks.named<Sync>("stonecutterGenerate").map { it.destinationDir })
-        resources.srcDirs(commonTasks.processResources.map { it.destinationDir })
+        java.srcDirs(extraJavaSources)
+        resources.srcDirs(extraResources)
     }
+}
+
+dependencies {
+    val commonPath = commonNode.project.path
+    // Manually depend on common's pre-processed sources.
+    // NOTE: loaders have no build dependency on common, so API/implementation
+    //  classes and transitive dependencies must be manually propagated.
+    extraJavaSources(project(path = commonPath, configuration = "generatedSourcesElements"))
+    extraResources(project(path = commonPath, configuration = "processedResourcesElements"))
 }
 
 val releaseMetadataTask = tasks.register<ProjectReleaseMetadataTask>("generateReleaseMetadata") {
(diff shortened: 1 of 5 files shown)


build-logic/conventions/src/main/kotlin/freecam.loaders.gradle.kts
```diff
@@ -4,14 +4,30 @@ plugins {
     id("freecam.common")
 }
 
+// TODO: do we need the binding, or do we get a DSL?
```
MattSturgeon: Needs addressing

build-logic/conventions/src/main/kotlin/freecam.loaders.gradle.kts
```diff
@@ -4,14 +4,30 @@ plugins {
...
+    isCanBeResolved = true
+}
+
+val commonProcessedResources = configurations.create("commonProcessedResources") {
+    isCanBeConsumed = false
+    isCanBeResolved = true
+}
+
+dependencies {
+    val commonPath = commonNode.project.path
```
MattSturgeon: Could be a top-level binding

common/build.gradle.kts
```diff
@@ -57,3 +57,24 @@ tasks.processResources {
...
+        add(name, tasks.stonecutterGenerate.map { it.destinationDir }) {
+            builtBy(tasks.stonecutterGenerate)
+        }
+    }
+}
+
+configurations.create("processedResourcesElements") {
+    isCanBeConsumed = true
+    isCanBeResolved = false
+    artifacts {
```
MattSturgeon, line 94: Inconsistent whitespace:
```suggestion

    artifacts {
```

common/build.gradle.kts
```diff
@@ -57,3 +57,24 @@ tasks.processResources {
 
     duplicatesStrategy = DuplicatesStrategy.FAIL
 }
+
+configurations.create("generatedSourcesElements") {
```
MattSturgeon, line 77: These consumable configurations probably warrant a `description`:


```suggestion
configurations.create("generatedSourcesElements") {
    description = "Consumable generated Java sources from the common project."
```

common/build.gradle.kts
```diff
@@ -57,3 +57,24 @@ tasks.processResources {
...
+    isCanBeResolved = false
+
+    artifacts {
+        add(name, tasks.stonecutterGenerate.map { it.destinationDir }) {
+            builtBy(tasks.stonecutterGenerate)
+        }
+    }
+}
+
+configurations.create("processedResourcesElements") {
```
MattSturgeon, line 89: same
```suggestion
configurations.create("processedResourcesElements") {
    description = "Consumable processed resources from the common project."
```

MattSturgeon (commented):
Before merging, we should test:
```console
$ ./gradlew clean buildAndCollect --configure-on-demand --no-build-cache
```
(repeat for `help`, `projects`, and `tasks`)

Let's add "build system cleanup and fixes" to the changelog, which maybe can be squashed with other changelog entries in the stable release.

PR #592, opened 2026-08-27
refactor(branding): convert SVG to monochrome evenodd path

MattSturgeon:
Although harder to maintain, due to lack of granular IDs and self-documenting shapes, a single `<path>` with `"evenodd"` fill-rule is mathematically simpler and has much wider support.

Extracted from #580, remove colours from the icon SVG, meaning the icon now only defines shape and inherits `currentColor` from context. This will simplify eventual re-use in (e.g.) the website.

### DIFF
--- a/branding/src/main/icon.svg
+++ b/branding/src/main/icon.svg
@@ -1,9 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <svg viewBox="0 0 512 512" xmlns="http://www.w3.org/2000/svg">
-  <rect id="background" width="512" height="512" fill="#002a36"/>
   <path
     id="camera"
-    fill="#00607c"
+    fill="currentColor"
     fill-rule="evenodd"
     d="
       M240 252
--- a/branding/src/main/icon.svg
+++ b/branding/src/main/icon.svg
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <svg viewBox="0 0 512 512" xmlns="http://www.w3.org/2000/svg">
   <path
-    id="camera"
     fill="currentColor"
     fill-rule="evenodd"
     d="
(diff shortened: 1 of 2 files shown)


branding/src/main/icon.svg
```diff
@@ -1,63 +1,62 @@
...
+      H270
+      A1 1 0 0 1 271 274
+      V285
+      A1 1 0 0 1 270 286
+      H252
+      V307
+      A1 1 0 0 1 251 308
+      H241
+      A1 1 0 0 1 240 307
+      Z
```
MattSturgeon, line 24: Central "F"-glyph, unchanged from the F-glyph that was in the mask.

branding/src/main/icon.svg
```diff
@@ -1,63 +1,62 @@
...
+
+      M312 278
+      A56 56 0 0 1 200 278
+      A56 56 0 0 1 312 278
+      Z
```
MattSturgeon, line 29: Outer lens circle, now declared as two semi-circle arcs

branding/src/main/icon.svg
```diff
@@ -1,63 +1,62 @@
...
+
+      M344 278
+      A88 88 0 0 1 168 278
+      A88 88 0 0 1 344 278
+      Z
```
MattSturgeon, line 34: Inner lens circle, also now two arcs

branding/src/main/icon.svg
```diff
@@ -1,63 +1,62 @@
...
+      M339 186
+      H415
+      A1 1 0 0 1 416 187
+      V207
+      A1 1 0 0 1 415 208
+      H339
+      A1 1 0 0 1 338 207
+      V187
+      A1 1 0 0 1 339 186
+      Z
```
MattSturgeon, line 45: Top-right flash rectangle, now declared as 4 lines and 4 1/4-circle arcs (rounded corners).

PR #598, opened 2026-08-30
fix: stop freecam from sinking when toggle sneak is enabled

reusteur73:
When Toggle Sneak is enabled (Options > Controls), the sneak
keybind becomes a toggle instead of a hold. Freecam's descend movement reads
that key's `isDown()` directly, so if sneak happened to be toggled on, the
camera would just sink forever with no way to stop it besides pressing sneak
again which also toggled the real player's crouch or by pressing jump key.

This makes freecam poll the sneak key's actual physical state instead,
independent of the toggle setting. Descending now works like holding jump to
ascend: only active while the key is actually held down.

Tested on Fabric 26.2. Compiles fine on the other supported versions too
(1.17.1 through 26.1, NeoForge 26.2), but I didn't run those in-game.


reusteur73:
Fixing

PR #599, opened 2026-08-31
fix: stop freecam from sinking when toggle sneak is enabled

reusteur73:
When Toggle Sneak is enabled (Options > Controls), the sneak
keybind becomes a toggle instead of a hold. Freecam's descend movement reads
that key's `isDown()` directly, so if sneak happened to be toggled on, the
camera would just sink forever with no way to stop it besides pressing sneak
again which also toggled the real player's crouch or by pressing jump key.

This makes freecam poll the sneak key's actual physical state instead,
independent of the toggle setting. Descending now works like holding jump to
ascend: only active while the key is actually held down.

Tested on Fabric 26.2. Compiles fine on the other supported versions too
(1.17.1 through 26.1, NeoForge 26.2), but I didn't run those in-game.

### DIFF
--- a/common/src/main/java/net/xolt/freecam/Freecam.java
+++ b/common/src/main/java/net/xolt/freecam/Freecam.java
@@ -3,6 +3,7 @@ package net.xolt.freecam;
 import net.minecraft.client.CameraType;
 import net.minecraft.client.KeyMapping;
 import net.minecraft.client.Minecraft;
+import net.minecraft.client.ToggleKeyMapping;
 import net.minecraft.client.multiplayer.ServerData;
 import net.minecraft.client.player.KeyboardInput;
 import net.minecraft.network.chat.Component;
@@ -41,6 +42,7 @@ public class Freecam {
     private static TripodSlot activeTripod = TripodSlot.NONE;
     private static FreeCamera freeCamera;
     private static CameraType rememberedF5 = null;
+    private static boolean sneakStateOnEnable = false;
 
     @ApiStatus.Internal
     public static void preTick(Minecraft mc) {
@@ -254,6 +256,7 @@ public class Freecam {
     private static void onEnable() {
         MC.smartCull = false;
         outlineEnabled = ModConfig.get().shouldOutlinePlayer();
+        sneakStateOnEnable = MC.options.keyShift.isDown();
 
         rememberedF5 = MC.options.getCameraType();
         //~ if >=26.2 getMainCamera -> mainCamera
@@ -279,6 +282,18 @@ public class Freecam {
     private static void onDisabled() {
         if (rememberedF5 != null) {
             MC.options.setCameraType(rememberedF5);
+        }
+        restoreSneakState();
+    }
+
+    // Using the sneak key to descend in freecam (see Motion.isSneakKeyDown) still reaches the real
+    // keyShift mapping, so with Toggle Sneak enabled it silently flips the player's persisted crouch
+    // state in the background. Flip it back if it changed while freecam was open, so freecam's own
+    // camera controls never leak into the player's actual sneak state once freecam closes.
+    private static void restoreSneakState() {
+        KeyMapping keyShift = MC.options.keyShift;
+        if (keyShift instanceof ToggleKeyMapping && keyShift.isDown() != sneakStateOnEnable) {
+            keyShift.setDown(true);
         }
     }
 
--- a/common/src/main/java/net/xolt/freecam/util/Motion.java
+++ b/common/src/main/java/net/xolt/freecam/util/Motion.java
@@ -1,7 +1,10 @@
 package net.xolt.freecam.util;
 
+import com.mojang.blaze3d.platform.InputConstants;
 import net.minecraft.util.Mth;
 import net.minecraft.world.phys.Vec3;
+
+import static net.xolt.freecam.Freecam.MC;
 
 public class Motion {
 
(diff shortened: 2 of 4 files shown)


common/src/main/java/net/xolt/freecam/mixins/KeyMappingAccessor.java
```diff
@@ -0,0 +1,13 @@
...
+import com.mojang.blaze3d.platform.InputConstants;
+import net.minecraft.client.KeyMapping;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+@Mixin(KeyMapping.class)
+public interface KeyMappingAccessor {
+
+    @Accessor("key")
+    InputConstants.Key freecam$getKey();
```
MattSturgeon: We can use [our `accessWidener`](https://github.com/MinecraftFreecam/Freecam/blob/main/common/src/main/resources/freecam.accesswidener) to grant access without mixins:

```
accessible field net/minecraft/client/KeyMapping key Lcom/mojang/blaze3d/platform/InputConstants$Key;
```

See [Fabric access widening docs](https://docs.fabricmc.net/develop/class-tweakers/access-widening)[^1]

[^1]: Note: we have a build step that converts our fabric access widener to a forge access transformer in the forge/neoforge builds.

common/src/main/java/net/xolt/freecam/util/Motion.java
```diff
@@ -50,10 +54,32 @@ public static void doMotion(FreeCamera freeCamera, double hSpeed, double vSpeed)
...
 
         freeCamera.setDeltaMovement(velocityX, velocityY, velocityZ);
     }
+
+    // The sneak keybind can be set to toggle rather than hold (Options > Controls > Toggle Sneak).
+    // In that case, MC.options.keyShift.isDown() reflects whatever sneak state the player toggled
+    // to beforehand, not whether the key is currently held, causing the camera to drift down for as
+    // long as that stale toggle happens to be on. Poll the physical key state directly instead, so
+    // descending in freecam always requires actually holding the key down, regardless of the toggle
+    // sneak setting.
```
MattSturgeon, line 68: As this is a user-facing change, can you please add a changeset file:

`.changeset/toggle-sneak.md`
```markdown
---
default: patch
---

# Stop freecam from sinking when toggle sneak is enabled
```

MattSturgeon (changes requested):
Thanks for the fix!

1.4.1 is currently on release candidate, so I'll land this after 1.4.1 has released.

In the mean time, could you switch from an Accessor Mixin to an Access Widener entry and add a change file to `.changeset`?

_(also re https://github.com/MinecraftFreecam/Freecam/pull/598, you don't need to close/open new PRs. It's perfectly fine to force-push to the existing PR)_

_(also, I've been doing some CI changes today, so you'll need to rebase on our latest `main` to run the correct 'required checks')_

MattSturgeon:
I took a few minutes to address my [earlier feedback](https://github.com/MinecraftFreecam/Freecam/pull/599#pullrequestreview-5067551094), so next time you work on this you'll need to overwrite your local branch with what I force-pushed to your GitHub fork.

<details><summary>Details</summary>
<p>

```shell
git stash
git fetch --all
git reset --hard @{u}
git stash pop
```

Note: `git reset --hard` reset what you have locally to what I pushed to your remote. That's probably what you want, unless you have made some changes locally that you didn't push and don't want to lose them.

Only run `git stash` and `git stash pop` if you have uncommitted changes you wish to re-apply after resetting to what I pushed.

</p>
</details>


---

I noticed in testing that using the sneak key while in freecam will cause toggle-sneak to togglne upon _exiting_ freecam. Ideally, we'd fix that too.

reusteur73:
Thanks for the feedback, I hadn't noticed that bug. I tried to implement a fix in `2cdef7a` and it seems to work properly. Let me know if you need further changes.

MattSturgeon:
~~The current CI failures are due to a maven repo being down: https://status.trans.fish/status/kiku (I'll trigger a re-run later).~~

Looking at the diff you pushed up, that _should_ work. Tracking state is never my preferred option, though. Perhaps it's possible for a mixin to prevent the key event propagating to the toggle-sneak binding while `Freecam.isEnabled()`, or something along those lines?

If that proves too messy or too fragile, the current solution should be ok.

common/src/main/java/net/xolt/freecam/Freecam.java
```diff
@@ -41,6 +42,7 @@ public class Freecam {
     private static TripodSlot activeTripod = TripodSlot.NONE;
     private static FreeCamera freeCamera;
     private static CameraType rememberedF5 = null;
+    private static boolean sneakStateOnEnable = false;
```
MattSturgeon, line 45: nit(style): Not super important, but maybe we should use the same style as `rememberedF5`:
```suggestion
    private static boolean rememberedSneakState = false;
```

Also: is there any value in making it nullable? I don't _think_ it is possible to get into a situation where we _disable_ freecam before the remembered state has been populated.
```java
private static Boolean rememberedSneakState = null;
```

common/src/main/java/net/xolt/freecam/Freecam.java
```diff
@@ -254,6 +256,7 @@ private static void onDisableFreecam() {
     private static void onEnable() {
         MC.smartCull = false;
         outlineEnabled = ModConfig.get().shouldOutlinePlayer();
+        sneakStateOnEnable = MC.options.keyShift.isDown();
```
MattSturgeon, line 259: If going down the "store state" approach, I think we'll also have to update it if the state changes during "control player" mode; Freecam has an option to toggle control of the player vs the camera while freecam is active.

MattSturgeon (commented):
My instinct is that there's a cleaner approach than storing/restoring remembered initial sneak state, likely involving a mixin.

I suspect we'll also need to give some consideration to the "freecam is active, but in 'control player' mode" scenario.

PR #548, opened 2026-07-22
Server policies

personalized-advertising:
The server can now define a set of policies for freecam mod users, i.e disabling clipping or freecam entirely, much more fine-grained than #547 as you requested.

You can implement this now and then if you later want to make a server-side mod its as simple as doing this from the server:

```java
public static final String FREECAM_POLICY_CHANNEL = "freecam:server_config";

public static void sendFreecamPolicy(Player player, boolean allowFreecam, boolean allowClipping, boolean allowFullbright, boolean allowInteract) {
    String json = """
            {
              "allowFreecam": %s,
              "allowClipping": %s,
              "allowFullbright": %s,
              "allowInteract": %s
            }
            """.formatted(
            allowFreecam,
            allowClipping,
            allowFullbright,
            allowInteract
    );
	
    player.sendPluginMessage(FREECAM_POLICY_CHANNEL, json.getBytes(StandardCharsets.UTF_8));
}
```

Untested, especially the forge code.

Issue #105, opened 2023-08-12
[Feature Request] Modrinth - Permissions


QarthO:
To follow Modrinth guidelines, you've had to made some adjustments to the features in order to upload to the site. I'm requesting in addition to requiring the user to be operator, or creative mode on a server, you also include a permission node. That way for servers who are okay with players in survival mode to use the full feature, they can still use the Modrinth version. This should also follow the guideline, as the server is providing the opt-in/opt-out permission node.

MattSturgeon:
Upon initial investigation, there appears to be multiple implementations targeted at different server platforms ([Bukkit](https://bukkit.fandom.com/wiki/CraftBukkit_Commands), [Spigot](https://www.spigotmc.org/wiki/spigot-commands), [Fabric](https://github.com/lucko/fabric-permissions-api)).

Do these permission systems report to the client what permissions are available? If so, how and how consistently with each other?

QarthO:
For Bukkit-based servers (spigot and paper will follow the same bukkit implementation) there will need to be a bukkit plugin made that will be the be the messenger to the client. It will supply the bukkit-permissions to the server, and when a player has the permission, the plugin will then tell the client.

For fabric, you would need to have the mod installed on the server doing the same thing.

https://luckperms.net/
This is the permissions plugin that basically every server uses, (bukkit/fabric/forge)

I've only worked on bukkit implementation, so not too sure how it would work on fabric, but can look at the implementations that other mods on bukkit servers send messages to the client such as..
https://modrinth.com/plugin/simple-voice-chat
https://modrinth.com/plugin/ping-wheel-plugin

So basically when the player joins, if they have permission to ghost thru blocks (which would be handled by the server plugin/mod), you'd [send a message]https://jd.papermc.io/paper/1.20/org/bukkit/plugin/messaging/PluginMessageRecipient.html#sendPluginMessage(org.bukkit.plugin.Plugin,java.lang.String,byte%5B%5D) to the client saying it is allowed. If the server either doesn't send a message, or the message says you cant, then you'd do nothing (the current implementation)

Hopefully this makes sense

MattSturgeon:
If bukkit/spigot servers _always_ sent a permissions packet automatically when joining a world it'd be a pretty simple feature. But needing to write & maintain an additional server-side mod/plugin makes this a little more effort.

Not sure @hashalite will be interested in doing that.

TBH if were writing server code, it begs the question why don't we just grant the player Spectator mode perms?

[...]

PR #633, opened 2026-09-22
feat: add server policies and AntiFreecam compatibility

TitusKirch:
## Summary

Allow servers to restrict Freecam features without changing players saved preferences. This builds on the approach in #548.

The motivation is to allow Freecam on our small private server while restricting features that could give players an unfair advantage.

## Changes

 - Receive JSON policies on `freecam:server_config`. Restrictions reset on disconnect; invalid updates preserve the current policy.
- Preserve AntiFreecam’s original behavior: `forceCollision=true` restricts clipping.
- Enforce restrictions on active cameras and interactions, including when switching player controls.
- Configure hosted-server policies through the settings menu or `serverPolicy` in the config file. LAN policy changes update connected clients.
- Keep unshared single-player worlds unaffected and allow clients without Freecam to join.

The branch also includes a build fix restricting KikuGie repository lookups to its own package namespace, plus ignore rules for generated IDE output.

## Validation

- CI passes for all targeted versions.
- All 16 supported loader/version combinations built locally.
- Configuration test executions passed.
- Dedicated-server startup and policy loading tested on Fabric 26.3, NeoForge 26.3, and Forge 1.17.1.
- Real Fabric 26.3 client/server testing.
- LAN multiplayer manually tested on Fabric 26.2.

## Scope

This provides feature restrictions, not a permission-node integration or a separate Bukkit/Paper plugin. Maximum camera distance, per-server local profiles, and a broader policy-notification system remain separate features.

## AI-assisted development

This implementation was developed using AI coding tools. I plan, descied, read and reviewed the changes and manually tested.

## Related work

Builds on #548 and the earlier work in #547.

Refs #546, #105.


.gitignore
MattSturgeon: I've cherry-picked and merged some of the "additional" changes. Can you rebase on the latest `main` and _drop_ your already-merged commits:
- `Merge pull request #2 from TitusKirch/fix/kikugie-repository-filter`
- `fix(build): restrict KikuGie repository lookups`
- `chore(gitignore): group Eclipse output under the correct heading`
- `chore(gitignore): ignore generated IDE output`

config/src/main/java/net/xolt/freecam/network/ServerPolicy.java
```diff
@@ -0,0 +1,17 @@
...
+    public static final ServerPolicy ALLOW_ALL = new ServerPolicy(true, true, true, true);
+
+    public String toJson() {
+        JsonObject json = new JsonObject();
+        json.addProperty("allowFreecam", allowFreecam);
+        json.addProperty("allowClipping", allowClipping);
+        json.addProperty("allowFullbright", allowFullbright);
+        json.addProperty("allowInteract", allowInteract);
+        return json.toString();
+    }
```
MattSturgeon: Why manually construct a JsonObject? Gson is able to serialize a java `record` automatically.

```java
gson.toJson(serverPolicyRecord)
// ServerPolicy record -> json string
```

config/src/test/kotlin/net/xolt/freecam/network/HostedServerPolicyTest.kt
```diff
@@ -0,0 +1,52 @@
...
+class HostedServerPolicyTest {
+    @AfterTest
+    fun cleanup() {
+        HostedServerPolicy.configure(ModConfigDTO())
+        ServerPolicies.reset()
+    }
+
+    @Test
+    fun `saved client configuration can be loaded as a server policy`() {
+        val dir = createTempDirectory("freecam-server-policy")
```
MattSturgeon: In other tests I've been using this pattern:

```suggestion
        val dir = createTempDirectory("freecam-server-policy").apply {
            toFile().deleteOnExit()
        }
```

which avoids the need for `try { ... } finally { deleteRecursively() }`

config/src/main/java/net/xolt/freecam/config/model/ServerPolicyConfig.java
```diff
@@ -0,0 +1,15 @@
+package net.xolt.freecam.config.model;
+
+import net.xolt.freecam.network.ServerPolicy;
+
+/** Settings for servers hosted by this installation, independent of received restrictions. */
+public class ServerPolicyConfig {
+    public boolean allowFreecam = true;
+    public boolean allowClipping = true;
```
MattSturgeon: Naming/terminology nit (applies to internal names and user-facing text):

Freecam doesn't usually use the term "clipping", our config options instead refer to "collision".

I'm wondering if we should flip the `allowClipping = true` to a `enforceCollision = false`, or find some other way to use consistent terminology.

We could also come up with something like `allowIgnoringCollision`, but that gets a bit wordy. It does set a precedent for more granular collision policies, though; e.g. we could add `allowIgnoringTransparentCollision` and `allowIgnoringOpenableCollision` options. Or maybe we could make collision options into a JSON object, instead of a single boolean...

TL;DR: I'd like consistent terminology, and I'd like to consider how the names/schema we choose now could affect future expansion to more granular policies.

config/src/main/java/net/xolt/freecam/config/model/ServerPolicyConfig.java
```diff
@@ -0,0 +1,15 @@
...
+    public boolean allowFullbright = true;
+    public boolean allowInteract = true;
+
+    public ServerPolicy snapshot() {
+        return new ServerPolicy(allowFreecam, allowClipping, allowFullbright, allowInteract);
+    }
```
MattSturgeon: Can we replace `snapshot` with implementing `java.lang.Cloneable`

fabric/build.gradle.kts
```diff
@@ -164,7 +165,8 @@ tasks {
 
             depends("minecraft", meta.reqs["mc"]?.toString() ?: error("${project.path} missing reqs.mc"))
             depends("fabricloader", meta.reqs["fabric_loader"]?.toString() ?: error("${project.path} missing reqs.fabric_loader"))
-            fabricApiModules.forEach { depends(it, "*") }
+            listOf("fabric-api-base", "fabric-lifecycle-events-v1", "fabric-networking-api-v1")
```
MattSturgeon: This should be reverted back to:
```kotlin
fabricApiModules.forEach { depends(it, "*") }
```

Was there a reason you wanted to exclude `fabric-resource-loader-v1` and `fabric-key-mapping-api-v1` from the declared dependencies?

i18n/src/main/en_US/config.toml
```diff
@@ -119,3 +119,11 @@ freecam.tooltip = "Notifies you when entering/exiting freecam."
...
 tripod.label = "Tripod Notifications"
 tripod.tooltip = "Notifies you when entering/exiting tripod cameras."
+
+[freecam.config.serverPolicy]
+label = "Hosted Server Policy"
+tooltip = "Restrictions for players in your LAN world or dedicated server. Does not change your personal settings."
+allowFreecam = "Allow Freecam"
+allowClipping = "Allow Clipping"
+allowFullbright = "Allow Full Brightness"
+allowInteract = "Allow Interactions in Freecam"
```
MattSturgeon: These options should have separate `label` and `tooltip` keys.

Also, again: I'd like to avoid the term "clipping".

README.md
```diff
@@ -84,6 +83,27 @@ See the [contributing guide](https://github.com/MinecraftFreecam/Freecam/blob/ma
 
 <!-- website:stop -->
 
+## Server policies
+
+Servers can temporarily restrict Freecam, clipping, full brightness, and interactions. Received rules reset on disconnect and never overwrite your saved preferences. Freecam also understands [AntiFreecam](https://github.com/Kesuaheli/AntiFreecam)'s collision restriction.
+
+For a LAN world, configure **Hosted Server Policy** in the settings menu, then open the world to LAN. Rules apply to compatible connected clients, including the host. Saving policy changes updates connected clients; unshared single-player worlds are unaffected.
```
MattSturgeon: > unshared single-player worlds are unaffected.

Why is that? The integrated server also runs on unshared singleplayer, so I'd assume it to behave the same as a hosted LAN world. Did you manually disable server policies when the LAN world is not open?

config/src/main/java/net/xolt/freecam/network/ServerPolicies.java
```diff
@@ -0,0 +1,113 @@
...
+import com.google.gson.stream.JsonToken;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.StringReader;
+import java.nio.charset.StandardCharsets;
+
+/** Connection-scoped overrides; never serialized into the local configuration. */
+public final class ServerPolicies {
```
MattSturgeon: I'm not sure how I feel about the policy state living in a static-singleton class, rather than (e.g.) an SPI loaded _instance_ of the class, like we do with some other things like config-loading (`ModConfig.get()`, etc).

I haven't thought too hard about this, but my instinct is that an instance-based model would be more flexible, make future expansion easier, make testing easier, etc.

Not blocking, but worth a discussion at least.

common/src/main/java/net/xolt/freecam/network/ServerPolicyManager.java
```diff
@@ -0,0 +1,62 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.IdentityHashMap;
+import java.util.Map;
+import java.util.function.BiPredicate;
+
+/** Server-side lifecycle; never loads client classes on a dedicated server. */
```
MattSturgeon: Should we be more explicit?

```suggestion
/** Server-side lifecycle; must not load client-side classes. */
```

cloth-config/src/main/java/net/xolt/freecam/config/gui/ModConfigScreenFactory.java
```diff
@@ -472,6 +473,32 @@ private SubCategoryListEntry notificationCategory(ConfigEntryBuilder entryBuilde
         return builder.build();
     }
 
+    private SubCategoryListEntry serverPolicyCategory(ConfigEntryBuilder entryBuilder) {
```
MattSturgeon: While I agree the config DTO (and JSON) should keep server policies separate, I think the user-facing GUI would probably be simpler if we combine the new policy settings into the existing `serversCategory`. Its user-facing label should already cover the expansion:

```toml
label = "Multiplayer Options"
tooltip = "Options relating to multiplayer servers."
```

Maybe the tooltip could remove "servers" to be more general? `Options relating to multiplayer`?

.changeset/server-policies.md
```diff
@@ -0,0 +1,11 @@
+---
+default: minor
+---
+
+Servers can temporarily restrict Freecam, clipping, full brightness, and interactions using a raw UTF-8 JSON plugin message on `freecam:server_config`.
+
+The boolean fields `allowFreecam`, `allowClipping`, `allowFullbright`, and `allowInteract` default to `true` when omitted. Each valid message replaces the previous policy; invalid messages leave it unchanged. Restrictions reset on disconnect and never change the saved local configuration.
+
+Supports AntiFreecam's `antifreecam:freecam_config_packet` boolean: `forceCollision=true` disables clipping, independently of JSON restrictions, and resets on disconnect.
```
MattSturgeon: This could probably be a separate change file, but I don't think the anti-freecam note needs much more detail than:

```suggestion
Added built-in support for AntiFreecam.
```

.changeset/server-policies.md
```diff
@@ -0,0 +1,11 @@
+---
+default: minor
+---
+
```
MattSturgeon: Change files need a h1 heading/summary. _Ideally_, that summary can fully-encapsulate the user-facing change, but if needed we can continue with more detail.

Something like:


```suggestion

# Freecam can now be installed server-side, to configure server-controlled policies

```

.changeset/server-policies.md
```diff
@@ -0,0 +1,11 @@
...
+Servers can temporarily restrict Freecam, clipping, full brightness, and interactions using a raw UTF-8 JSON plugin message on `freecam:server_config`.
+
+The boolean fields `allowFreecam`, `allowClipping`, `allowFullbright`, and `allowInteract` default to `true` when omitted. Each valid message replaces the previous policy; invalid messages leave it unchanged. Restrictions reset on disconnect and never change the saved local configuration.
```
MattSturgeon: I'd like to avoid fully-documenting the feature in the changelog. The changelog should just tell users "a thing changed" or "a thing was added", with a bit of context.

That said, I don't mind including detail if it is relevant and can be skimmed over. Maybe something like

```md
Added initial server policies: `allowFreecam`, `allowClipping`, `allowFullbright`, and `allowInteract`.
Servers can configure these policies to temporarily restrict Freecam's features while connected to that server.

Users can configure these policies for open-to-LAN worlds in Freecam's "Multiplayer Options".
Servers can install Freecam and use the same config file to enforce policies.

See our [README](https://github.com/MinecraftFreecam/Freecam/?tab=readme-ov-file) for more detail.
```
?

common/src/main/java/net/xolt/freecam/config/model/ModConfigImpl.java
```diff
@@ -34,17 +35,17 @@ public double getVerticalSpeed() {
 
     @Override
     public boolean ignoreAllCollision() {
-        return data.collision.ignoreAll;
+        return ServerPolicies.allowClipping() && data.collision.ignoreAll;
```
MattSturgeon: One limitation of integrating this directly into `ModConfig` is that calling code does not know whether it was the user's config or the server's policy that disabled the feature.

If the calling code knew it was a server-side restriction, it could display a status message informing the user why their action failed (like we already do with enabling freecam on a blacklisted server URL).

MattSturgeon (commented):
Initial review pass. Looking good overall, I've left some comments below.

I didn't have time to test this out in any detail or look through in the IDE, so don't be offended if I find more things to comment on in a future review.

Thanks for working on this!!

fabric/build.gradle.kts
TitusKirch: I assumed the client-only key mapping module would break dedicated servers, but testing shows Fabric Loader treats environment-disabled mods as satisfied dependencies and the server starts fine. I revert this.

config/src/main/java/net/xolt/freecam/config/model/ServerPolicyConfig.java
TitusKirch: I agree. I've renamed it and switched to a nested JSON object so it can be extended later. I've left the granular options out for now. IMO they'd be better tracked in a separate issue, with this PR providing the base schema for them. Unknown keys inside collision are already ignored, so adding them later won't break older clients.

config/src/main/java/net/xolt/freecam/network/ServerPolicies.java
TitusKirch: Agreed, an instance-based model would be cleaner and would make the tests simpler. It also overlaps with your comment on ModConfigImpl. I'd suggest tackling both together in a follow-up issue to keep this PR small, unless you'd prefer it here.

common/src/main/java/net/xolt/freecam/config/model/ModConfigImpl.java
TitusKirch: Agreed, but I'd handle it in the same follow-up issue as the ServerPolicies refactor

README.md
TitusKirch: The idea was that the policy is a ruleset for other players. Someone who restricts freecam for their LAN guests probably doesn't expect to be restricted in their own solo worlds too, where their personal config already decides what they can do. Once the world is opened to LAN, it applies to everyone, including the host.

config/src/main/java/net/xolt/freecam/config/model/ServerPolicyConfig.java
TitusKirch: `clone()` would be a shallow copy sharing the nested collision object and lose the record's equals. How about replacing `snapshot()` with `ServerPolicy.create(config)` instead, following the `CollisionPredicate.create(...)` pattern?

TitusKirch:
> Initial review pass. Looking good overall, I've left some comments below.
> 
> I didn't have time to test this out in any detail or look through in the IDE, so don't be offended if I find more things to comment on in a future review.
> 
> Thanks for working on this!!

Thanks for the review! I've addressed most of the comments and left a few open for discussion. Happy to get more feedback and it's not entirely selfless either, I'd really like to see this feature land 😄

config/src/main/java/net/xolt/freecam/config/model/ServerPolicyConfig.java
MattSturgeon: > `clone()` would be a shallow copy

Does it have to be? I assumed we could implement `clone()` however we like, so we could use a deep-clone implementation if we wanted

README.md
MattSturgeon: For me it seems surprising that the host's behaviour is different depending on whether a world is currently open-to-LAN.

I'd expect locally configured server policies to _either_:
1. Always apply to the host
2. Never apply to the host

Both of these make sense to me, and if there was a need we could even introduce a "also apply these policies to me" option (in the future or in this PR).

config/src/main/java/net/xolt/freecam/config/model/ServerPolicyConfig.java
MattSturgeon: Oops, I glossed over how this is constructing a `ServerPolicy` from a `ServerPolicyConfig`, so we don't want a clone of the config.

Since we're constructing a new instance, the factory-method (`create(from)`) pattern could make sense, or we could rename `snapshot` to something like `toServerPolicy` or `asServerPolicy`?

fabric/src/main/java/net/xolt/freecam/fabric/FabricServerPolicyNetworking.java
```diff
@@ -0,0 +1,42 @@
...
+    public static void register() {
+        //? if >=1.20.5 {
+        ClientPlayNetworking.registerGlobalReceiver(
+                ServerPolicyPayload.TYPE,
+                (payload, context) -> ServerPolicies.applyJson(payload.json())
+        );
+        ClientConfigurationNetworking.registerGlobalReceiver(AntiFreecamPayload.TYPE,
+                (payload, context) -> ServerPolicies.applyAntiFreecam(payload.forceCollision()));
+        ClientPlayNetworking.registerGlobalReceiver(AntiFreecamPayload.TYPE,
+                (payload, context) -> ServerPolicies.applyAntiFreecam(payload.forceCollision()));
```
MattSturgeon: Have we tested what happens if both Freecam and AntiFreecam are installed together client-side, and both attempt to handle this packet?

I'm assuming we should just mark Freecam as incompatible with antifreecam going forward. I.e. AntiFreecam should only be installed alongside _older_ Freecam versions, that don't declare an incompatibility and don't implement AntiFreecam support.

Although fabric conflicts are not sided, that's not an issue because there should never be a reason to install both Freecam and AntiFreecam serverside anyway.

cc AntiFreecam maintainer: @Kesuaheli

config/src/main/java/net/xolt/freecam/config/model/ServerPolicyConfig.java
TitusKirch: I've migrated it to `ServerPolicy.create(config)` as proposed.

README.md
TitusKirch: I like the option idea. I'd add an "apply to me" option (default off).

fabric/src/main/java/net/xolt/freecam/fabric/FabricServerPolicyNetworking.java
TitusKirch: Tested it: with both installed, the game crashes on startup (`Packet type … is already registered!`). I've added breaks: antifreecam, so Fabric now shows a clean `incompatible mods` error instead.

TitusKirch:
I've worked through your feedback and pushed the changes. The only open threads are the ServerPolicies instance-based model and telling the user when the server restricted something. If I see it correctly, both need the same API change, so I'd suggest handling them together in a follow-up issue. Granular collision policies (e.g. `allowIgnoringTransparent`) would be another follow-up, since the collision object is already the base for them.

common/src/main/java/net/xolt/freecam/mixins/MinecraftMixin.java
```diff
@@ -28,7 +28,15 @@ private void onDoAttack(CallbackInfoReturnable<Boolean> ci) {
...
     }
 
-    // Prevents item pick when allowInteract is disabled.
+    // Prevents item use when interactions are disabled.
+    @Inject(method = "startUseItem", at = @At("HEAD"), cancellable = true)
+    private void onUseItem(CallbackInfo ci) {
+        if (freecam$disableInteract()) {
+            ci.cancel();
+        }
+    }
```
hashalite: This prevents eating in Freecam when the local `allowInteract` setting is disabled, which seems out of scope for this PR.

This change can be discussed further, but can we keep the focus of this PR on setting enforcement and preserve existing behavior for now?

common/src/main/java/net/xolt/freecam/mixins/MultiPlayerGameModeMixin.java
```diff
@@ -62,16 +61,16 @@ private void onInteractEntityAtLocation(Player player, Entity entity, EntityHitR
     }
     *///? }
 
-    // Prevents attacking self.
+    // Prevents attacking self or bypassing interaction restrictions.
     @Inject(method = "attack", at = @At("HEAD"), cancellable = true)
     private void onAttackEntity(Player player, Entity target, CallbackInfo ci) {
-        if (target == MC.player) {
+        if (target == MC.player || freecam$disableInteract()) {
```
hashalite: Is this `freecam$disableInteract()` check necessary here? We already block attacks in `MinecraftMixin` via `startAttack`, and it also seems out of scope.

common/src/main/java/net/xolt/freecam/Freecam.java
```diff
@@ -371,11 +388,22 @@ public static boolean isPlayerControlEnabled() {
...
     }
 
+    @ApiStatus.Internal
+    public static boolean shouldPreventInteractions() {
+        if (!isEnabled()) {
+            return false;
+        }
+        if (!ServerPolicies.allowInteract()) {
+            return true;
+        }
```
hashalite: Putting the `ServerPolicies.allowInteract()` check here disables interactions even when Player Control is active, bypassing our existing exception for that mode. To keep behavior consistent, the server policy should respect` isPlayerControlEnabled()` like the local config does. If we want a way for servers to disable Player Control interactions, that feels like a separate feature request/policy setting.

MattSturgeon:
Awesome thanks. I'm currently away from my main PC until next week, so I won't be able to get this landed immediately. My plan is to dedicate some time to play around with the changes locally, and review in more detail with an IDE.

Ill probably keep the next release (currently in alpha) dedicated to the 26.3 bump and small fixes, then target this feature for the following release.

@hashilite mentioned to me that he's also supportive of this PR, but that he'd spotted a few behavioural changes he'd like to discuss. Ideally, he'll post a review soon, otherwise I'll post something on his behalf when I'm next able to review.

With regards to the global singleton-class vs holder+instance model, my preference would be to get that right in the initial PR rather than needing to address it in a follow-up. But as I said initially, I won't block on it.

hashalite (commented):
Thanks for the PR! I've taken some time to look it over, and I'm happy with the direction it's going in. There are a few things I wanted to comment on, which I've included below.

fabric/src/main/java/net/xolt/freecam/fabric/FabricServerPolicyNetworking.java
MattSturgeon: Great thanks. For the upstream AntiFreecam repo they'll probably want to change their wording of:

> From now on, all players* that want to join, are required to have AntiFreecam installed on their client too.

after this is released, to something like:

> From now on, all players* that want to join, are required to have either AntiFreecam __or a recent version of Freecam that implements AntiFreecam__ installed on their client too.

Or perhaps AntiFreecam will be archived, if all use cases are covered by our "server policies".

.changeset/antifreecam.md
```diff
@@ -0,0 +1,5 @@
+---
+default: minor
+---
+
+# Added built-in support for AntiFreecam
```
MattSturgeon: Due to the client-side incompatiblity, I think we should note something like:

```suggestion
# Added built-in support for AntiFreecam

Freecam now natively supports servers with [AntiFreecam]() installed.
You should no longer install both Freecam and AntiFreecam.
```

.changeset/server-policies.md
```diff
@@ -0,0 +1,13 @@
+---
+default: minor
+---
+
+# Freecam can now be installed server-side, to configure server-controlled policies
```
MattSturgeon:
```suggestion
# Freecam can now be installed server-side, enabling server-controlled usage policies
```

README.md
```diff
@@ -10,8 +10,7 @@ It is your responsibility to check **all** relevant rules **before** using this
 
 <!-- website:stop -->
 
-While Freecam should only affect the client, it **may be possible** for servers to detect the mod being installed through various exploits in Minecraft,
-modding platforms, or this mod.  
+Freecam advertises networking channels so servers can send temporary restrictions. Servers may also detect the mod through other Minecraft or mod-loader behavior.
 Here's a [real example](https://github.com/MinecraftFreecam/Freecam/issues/196) that exploited translation keys.
```
MattSturgeon: Good catch updating this README note!

Do we still need this example exploit link now that Freecam is more explicitly detectable and configurable by servers? I believe your new note fully covers the topic.


```suggestion
```

README.md
```diff
@@ -84,6 +83,29 @@ See the [contributing guide](https://github.com/MinecraftFreecam/Freecam/blob/ma
...
+    "collision": {
+      "allowIgnoring": false
+    },
+    "allowFullbright": false,
+    "allowInteract": true
+  }
+}
+```
+
+Clients without Freecam can still join. A custom server plugin can instead send the same fields as a raw UTF-8 JSON object on `freecam:server_config` (without the `serverPolicy` wrapper). Each valid message replaces the previous JSON policy; omitted fields allow the feature. AntiFreecam's separate restriction still applies until it is lifted or the client disconnects.
```
MattSturgeon:
```suggestion
Clients without Freecam can still join and clients with a different implementation of freecam are unaffected.

Instead of installing Freecam on your server, a custom server plugin can send a compatible UTF-8 JSON payload on the `freecam:server_config` channel. The payload schema is the same JSON that's saved in a `config/freecam.json` config, without the outer `serverPolicy` wrapper. Each valid message replaces the previous policy; omitted fields allow the feature.
```

README.md
```diff
@@ -84,6 +83,29 @@ See the [contributing guide](https://github.com/MinecraftFreecam/Freecam/blob/ma
 
 <!-- website:stop -->
 
+## Server policies
+
+Servers can temporarily restrict Freecam, ignoring collision, full brightness, and interactions. Received rules reset on disconnect and never overwrite your saved preferences. Freecam also understands [AntiFreecam](https://github.com/Kesuaheli/AntiFreecam)'s collision restriction.
+
+For a LAN world, configure the policy options under **Multiplayer Options** in the settings menu, then open the world to LAN. Rules apply to compatible connected clients; saving policy changes updates them. The host is not restricted by default. Enable **Apply Policies to Me** to restrict yourself too, in every world you host, whether or not it is open to LAN.
```
MattSturgeon:
```suggestion
For singleplayer, configuring the policy options under **Multiplayer Options** in the settings menu will affect players connected via open-to-LAN. Policies apply to compatible connected clients; saving policy changes updates them.
By default, you are not affected by your own policies: enable **Apply Policies to Me** to also restrict yourself in singleplayer worlds.
```

README.md
```diff
@@ -84,6 +83,29 @@ See the [contributing guide](https://github.com/MinecraftFreecam/Freecam/blob/ma
 
 <!-- website:stop -->
 
+## Server policies
+
+Servers can temporarily restrict Freecam, ignoring collision, full brightness, and interactions. Received rules reset on disconnect and never overwrite your saved preferences. Freecam also understands [AntiFreecam](https://github.com/Kesuaheli/AntiFreecam)'s collision restriction.
```
MattSturgeon:
```suggestion
Servers can temporarily restrict Freecam, ignoring collision, full brightness, and interactions. Received rules reset on disconnect and never overwrite your saved preferences. Freecam also supports servers using [AntiFreecam](https://github.com/Kesuaheli/AntiFreecam).
```

cloth-config/src/main/java/net/xolt/freecam/config/gui/ModConfigScreenFactory.java
```diff
@@ -429,11 +429,66 @@ private SubCategoryListEntry serversCategory(ConfigEntryBuilder entryBuilder) {
                 .setSaveConsumer(value -> config().servers.blacklist = value)
                 .build();
```
MattSturgeon: IIRC cloth-config has a "label" entry for adding text without an actual option.

I think it may be useful to add something that calls out

> These options restrict Freecam usage by other players connecting to your singleplayer worlds over LAN.
> This only affects players using a compatible Freecam mod, such as this one.

This explicitly makes users aware of the scope of the options:
- Affects other players
  - unless "also apply to me", but that's self-evident
- Applied to singleplayer worlds
  - Doesn't apply when joining a multiplayer world

This could be a good place to centralise notes currently repeated in tooltips, too?

i18n/src/main/en_US/config.toml
```diff
@@ -119,3 +119,22 @@ freecam.tooltip = "Notifies you when entering/exiting freecam."
...
+allowIgnoringCollision.tooltip = "Whether players can travel through blocks in freecam."
+
+allowFullbright.label = "Allow Full Brightness"
+allowFullbright.tooltip = "Whether players can use full brightness in freecam."
+
+allowInteract.label = "Allow Interactions"
+allowInteract.tooltip = "Whether players can interact with blocks/entities in freecam."
+
+applyToHost.label = "Apply Policies to Me"
+applyToHost."tooltip[0]" = "Whether the policies above also restrict you in worlds you host."
```
MattSturgeon:
```suggestion
applyToHost."tooltip[0]" = "Whether the policies above also restrict you in singleplayer worlds."
```

"Host" somewhat implies that a server is being hosted for multiplayer access. Technically, that's true (Minecraft always hosts a server, even if it's not "open" to external connections), however most players are probably more used to thinking of "worlds they host" as "worlds started from the 'singleplayer' menu". I think that better encapsulates the "regardless of open-to-LAN" model.

MattSturgeon (commented):
I did a pass on user-facing text. On my phone so may have made mistakes, happy to discuss.

Key terminology change was introducing "singleplayer worlds" to imply "any world opened through the singleplayer menu, regardless of it's open-to-LAN state".

I suppose even that could be confusing; someone might think "I'm joining my friend's singleplayer world that they opened to LAN", however to do that they must use the multiplayer menu.

common/src/main/java/net/xolt/freecam/mixins/MinecraftMixin.java
MattSturgeon: My view on "allow interact" from a server-policy perspective is subtly different to a client-config perspective.

Client-side, I may just want to disable interactions outright for convenience. But from a server-policy perspective, I should only care if the player is able to do thi9they otherwise couldn't; i.e. interactions from the camera's perspective.

All other interactions (those from the player's perspective, and those that have no frame of reference like eating) should not affect multiplayer gameplay, as they happen from the actual player's perspective and are no different to e.g. quickly exiting Freecam, interacting, and returning to Freecam.

common/src/main/java/net/xolt/freecam/Freecam.java
MattSturgeon: https://github.com/MinecraftFreecam/Freecam/pull/633#discussion_r4099529706 also applied here, perhaps more so

common/src/main/java/net/xolt/freecam/mixins/MinecraftMixin.java
MattSturgeon: Thinking about it, we should probably drop the `allowInteract` server policy and replace it with a `allowInteractionsFromCamera` or `allowCameraInteractions` or something. That avoids hinting to servers that they may wish to block _any_ form of allowing interactions.

PR #634, opened 2026-09-23
chore: bump fabric-loom to 1.18

MattSturgeon:


MattSturgeon:
> Dependency requires at least JVM runtime version 25. This build uses a Java 21 JVM.

Looks like we need to bump the Gradle JVM version to use loom 1.18. Not sure why it worked locally, maybe I already had that configured in my IDE (but not CI).

